PR #97, opened 2015-07-15
ScalaJS port

rmihael:
Hi.

I want to propose this cross-port of `squants` for ScalaJS. In my practice I found `squants` to be very useful for domain modeling, but lack of transparency between Scala and ScalaJS always complicated things for me, forcing to provide `squants`-free wrappers and other inconvenient workarounds. Finally I decided to put some efforts to eliminate this problem.

Summary of changes:
1. Merged `build.sbt` into `Build.scala`. It's not always clear how Oh-So-Simple Build Tool will merge these files in case of multi-project setup, so single `Build.scala` is potentially less troublesome.
2. Removed dependency on `spire`. It lacks ScalaJS-compatible artifact and was only used in single test for minor utility purpose. I updated this test to work without `spire`.
3. Minor update in strings parsing regexp to be compatible with JS behavior. From my standpoint this difference in behavior should be treated as ScalaJS bug and I'll create a ticket for it, but my fix is really straightforward so it should not constitute a problem.
4. Minor update in formatting logic to work uniformly between JVM and JS.
5. Most unpleasant part: ScalaJS's `CrossProject` is not perfect yet, so in order to get everything working property I had to move `src` folder into `shared/src`. That's a major reason why PR looks so big.
6. I had to update some support libraries to their most recent versions. `scalatest` is most notable as I had to use milestone artifact for it. I assume it's not a problem for a `SNAPSHOT` version.
7. Some minor changes in TravicCI configuration. I found Travis cache to be very useful in my practice, so I enabled it.

As result, just by running `test`, `compile` or `publish...` in SBT console you'll have it run both for JVM and JS builds of `squants`. It's possible to narrow it down to specific build by choosing corresponding project through `project squantsJVM` or `project squantsJS`.

P.S. Let me just say that I admire quality of your tests coverage. It's really enabling for a people like me to propose some non-trivial changes.

[Review on Reviewable](https://reviewable.io/reviews/garykeorkunian/squants/97)


antonkulaga:
Great job, @rmihael 
I also need squants on the client a lot!

antonkulaga:
Any feedback about this PR? ScalaJS support is very important feature

garyKeorkunian:
Hey all, sorry I missed this when it was first posted.  

Squants on the client side is definitely needed.  I'll begin reviewing this PR today.  Hopefully I have it merged (or some comments) soon.

Thanks for your contribution.

garyKeorkunian:
This PR has been merged, but unfortunately broke the sbt-console and publish features - https://github.com/garyKeorkunian/squants/issues/101

I am working on a fix now.  Any tips on how to quickly resolve this would be great.  I suspect this may have be caused by the move from the src -> shared directories.

antonkulaga:
@garyKeorkunian 
console imports were applied to squantsJS and squantsJVM (because of CrossProject helper that applied default settigs to both of them) but not to RootProject as root was created automaticly by sbt, if you want the console to work for root, apply default settings to root project.

garyKeorkunian:
@antonkulaga @rmihael 

Ah, yes, thanks.

Appears I can use console and publish by first switching to the squantsJVM project.

Not sure exactly what I need to do to get that to work from root.

Also, what's the best way to make the JS library available to that community?

antonkulaga:
@garyKeorkunian ScalaJS version off the library is published in a same way as JVM version. I just automaticly adds sjs suffix to the artifact name, everything else is the same, see http://www.scala-js.org/doc/sbt/depending.html for details

PR #98, opened 2015-07-21
storage support (bytes)

derekmorr:
This is basic support for computer storage (bytes). Supports both base 10 and base 2 units.

### DIFF
--- a/src/main/scala/squants/storage/Storage.scala
+++ b/src/main/scala/squants/storage/Storage.scala
@@ -1,0 +1,214 @@
+/*                                                                      *\
+** Squants                                                              **
+**                                                                      **
+** Scala Quantities and Units of Measure Library and DSL                **
+** (c) 2013-2015, Gary Keorkunian                                       **
+**                                                                      **
+\*                                                                      */
+
+package squants.storage
+
+import squants._
+
+/**
+ * Represents computer storage.
+ *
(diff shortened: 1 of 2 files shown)


derekmorr:
Any chance I could get this reviewed?

garyKeorkunian:
Sorry, I missed this when it first came through.

I'll take a look at it today.

Thanks for your submission.

onilton:
Didn't know the right place to ask this, but I wonder why the string representation doesn't include the B as in the common unit symbols (i.e. `GiB` instead of `Gi`) ? 

https://www.wikiwand.com/en/Gibibyte

garyKeorkunian:
Good question.  @derekmorr any reason for excluding the B?

derekmorr:
The print section of the Unified Code of Units of Measure listed them without the B.

The site seems down at the moment, but the Google cache link is https://webcache.googleusercontent.com/search?q=cache:PFXiSDcy1tsJ:unitsofmeasure.org/ucum.html+&cd=1&hl=en&ct=clnk&gl=us (see section 4.6).

But I agree that it would be more useful to include the B. I'll send an updated PR with it.

Also, I think it would be better to rename this as "Information" instead of "Storage" since most standards list bytes as the unit of measure for information. What do you think ?

PR #105, opened 2015-09-03
Addition of unit KilometersPerSecond to Velocity.

fnussber:
I am working for [Gemini Observatory](http://www.gemini.edu) and we are in the process of adopting squants for parts of our [Observatory Control System](https://github.com/gemini-hlsw/ocs) to gain additional type safety and to be able to represent and persist user inputs as values with a specific unit (as chosen by the users). Thank you very much for your work!

Dealing with speeds I was 
* missing a unit to represent kilometers per second,
* noted that the symbol for kilometers per hour is "km/s" instead of "km/h" and
* noted that `kilometerPerSecond` in `VelocityConversions` is defined as `KilogramsPerSecond(1)`.

Here is a PR for your consideration that contains a change to the symbol for kilometers per hour, changes the definition of `kilometerPerSecond` and adds a unit for kilometers per second.

### DIFF
--- a/shared/src/main/scala/squants/motion/Velocity.scala
+++ b/shared/src/main/scala/squants/motion/Velocity.scala
@@ -41,6 +41,7 @@ final class Velocity private (val value: Double, val unit: VelocityUnit)
 
   def toFeetPerSecond = to(FeetPerSecond)
   def toMetersPerSecond = to(MetersPerSecond)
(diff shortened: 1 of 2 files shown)


garyKeorkunian:
Good catch on the unit and conversion issues.  Thanks for you contribution.

PR #108, opened 2015-09-18
Allow numbers with exponential notation when parsing a Dimension.

fnussber:
I noticed that currently the parsing of `Dimension`s does not support exponential notation like e.g. `5.0e-17` etc. Executing `Velocity("2.0e3 m/s")` for example will fail.

Here is a PR for your consideration that contains a change to the regular expression used when splitting a string into the number and units part and a few additional tests.


garyKeorkunian:
Nice addition.  Thanks!

PR #115, opened 2016-02-03
Fixes to AreaDensity

DaveDeCaprio:
We are using Squants at Cibo Technologies for units in agricultural applications.  We needed some new AreaDensity types, but the conversions in the base AreaDensity assumed that the type was KilogramsPerSquareMeters.

I've added two new AreaDensity units and fixed the base class so it works with multiple units.


garyKeorkunian:
Dave, this looks great.

One thing before I merge the PR ... can you fix the indentation on the lines you added so they match with the rest of the code.  I know it seems a bit nit-picky, but it helps keep things clean.

Thanks for your contribution!!

DaveDeCaprio:
Fixed.  My editor is set to tabs = 2 so I didn't notice.  All good now.

DaveDeCaprio:
My editor was set to tabs=2 so I didn't notice.  All fixed.

PR #117, opened 2016-02-09
Link scaladocs to their sources on Github

joshlemer:
[Here](http://i.imgur.com/98aSUFu.png) is an example of what it looks like.


joshlemer:
@garyKeorkunian Can we get a rebuild? The error was just a timeout.

PR #125, opened 2016-04-23
Spectral irradiance

cquiroz:
This PR adds the quantity `SpectralIrradiance` and their related units. Unit tests are provided for the quantity and units

[SpectralIrradiance](https://en.wikipedia.org/wiki/Irradiance#Spectral_irradiance) is used in astronomy for a variety of measurements

### DIFF
--- a/shared/src/main/scala/squants/radio/Irradiance.scala
+++ b/shared/src/main/scala/squants/radio/Irradiance.scala
@@ -9,8 +9,8 @@
 package squants.radio
 
 import squants._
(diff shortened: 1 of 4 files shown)


cquiroz:
This PR was originally made by @fnussber ande we are using it in the [ocs](https://github.com/gemini-hlsw/ocs) project. I added unit tests and send it to you for checks.

It would be useful for us to have it integrated into squants rather than having our own separate version

Cheers

Issue #127, opened 2016-05-08
Provide information on type level if e.g. one dimension is the product of two other dimensions


nightscape:
Hi @garyKeorkunian,

I would like to use squants in a type-class based setting where it is important to keep track on a type level if e.g. two values of two different types can be multiplied to get a value of a third type.
I would like to derive something like the `CanMultiply` class in the following (hugely simplified) example:
```scala
implicit def createMultiplication[InDim1, InDim2, OutDim](implicit canMultiply: CanMultiply[InDim1, InDim2, OutDim]) = new Multiplication[InDim1, InDim2, OutDim](...)

val areaMultiplication: Multiplication[Length, Length, Area] = implicitly
val areaDensityMultiplication: Multiplication[Density, Length, AreaDensity] = implicitly
```

Skimming through the code I haven't found any hint that dimensions are actually tracked on a type level.
Is something like this planned?
Would you accept a PR if I find the time to create one?

nightscape:
Something like this could be used to track the exponents of the used dimensions on a type level:
https://gist.github.com/non/8396462

PR #128, opened 2016-05-08
Tracking dimensions at type level

nightscape:
This PR is used to keep track of an implementation of #127.
It is not yet meant to be merged.
The current state just works for a single "base dimension" like `Length` where it can prove that 

```
IsProductOf[Area, Length, Length]
IsProductOf[Volume, Area, Length]
IsProductOf[Volume, Length, Area]
```

### DIFF
--- a/shared/src/main/scala/squants/DimensionOrder.scala
+++ b/shared/src/main/scala/squants/DimensionOrder.scala
@@ -7,8 +7,8 @@ trait DimensionOrder[D] {
 }
 object DimensionOrder {
   type Aux[D, Index] = DimensionOrder[D] { type I = Index }
-  implicit val lengthIs1: Aux[Length, _1] = null
-  implicit val massIs2: Aux[Mass, _2] = null
-  implicit val timeIs3: Aux[Time, _3] = null
-  implicit val energyIs4: Aux[Energy, _4] = null
+  implicit val lengthIs1: DimensionOrder[Length] { type I = _1 } = null
+  implicit val massIs2: Dimension[Mass] { type I = _2 } = null
+  implicit val timeIs3: Dimension[Time] { type I = _3 } = null
+  implicit val energyIs4: Dimension[Energy] { type I = _4 } = null
 }
--- a/shared/src/main/scala/squants/DimensionOrder.scala
+++ b/shared/src/main/scala/squants/DimensionOrder.scala
@@ -3,12 +3,41 @@ package squants
 import squants.TypeLevelInt._
 
 trait DimensionOrder[D] {
-  type I <: Pos
+  type DimensionIndex <: TypeLevelInt
 }
+
+trait DimensionIsEarlierThan[D1, D2]
+
+
 object DimensionOrder {
-  type Aux[D, Index] = DimensionOrder[D] { type I = Index }
-  implicit val lengthIs1: DimensionOrder[Length] { type I = _1 } = null
-  implicit val massIs2: Dimension[Mass] { type I = _2 } = null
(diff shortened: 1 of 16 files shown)


nightscape:
Ok, this took me quite some time to figure out and I've learnt more about the Scala type system (including the incredible SingletonOf hack) than in the last 3 years together :smile:
It is now possible to do the following:
```scala
implicitly[Volume IsProductOf (Length, Area)]
implicitly[AreaDensity IsProductOf (Density, Length)]
implicitly[Mass IsProductOf (Area, AreaDensity)]
implicitly[Length IsProductOf (Velocity, Time)]
```
The remaining caveats are that

1. the dimensions have to be [coded as an HList](https://github.com/nightscape/squants/blob/typelevel_dimensions/shared/src/main/scala/squants/mass/Mass.scala#L31) (I copied [the code from shapeless](https://github.com/nightscape/squants/blob/typelevel_dimensions/shared/src/main/scala/squants/HList.scala) for the time being because I didn't know if you're willing to add this as an external dependency) and
2. that you have to [explicitly specify an order](https://github.com/nightscape/squants/blob/typelevel_dimensions/shared/src/main/scala/squants/DimensionOrder.scala#L26-L29) and [use the same order in the quantity class definitions](https://github.com/nightscape/squants/blob/typelevel_dimensions/shared/src/main/scala/squants/mass/AreaDensity.scala#L22).

Nonetheless, I think this should help a lot for use cases where you want to make sure that certain operations are type-safe or in a type-class based approach where you might need the the type information for the compiler to work correctly.

If you had the time to look at the changes, we can talk about if and how to merge this PR (personally I like to rebase and cleanup git history before merging stuff).

garyKeorkunian:
Cool stuff.  At Scala Days now, but will be reviewing this over the next several days.  

Thanks,

nightscape:
Ok, have fun at ScalaDays! Looking forward to videos and slides :+1:

nightscape:
A technique like the following could be used to get dimensions into the right order:
https://milessabin.com/blog/2012/01/27/type-level-sorting-in-shapeless/

ngbinh:
any progress on this? Thanks

nightscape:
From my point of view this could be merged. It doesn't work in all cases and maybe someone comes up with a better solution of how to encode the dimensions which doesn't require them to be ordered, but it shouldn't break any existing functionality and might already be useful to some people.

cquiroz:
I don't have any strong opinion about this PR. Should we include it on the next release?

golem131:
@cquiroz i think, we must review this PR more accurate. May be plan it on 1.1 (or 1.0.1) milestone?

vitorsvieira:
Me neither, but something brought up my attention. This PR adds a copy of HList implementation from Miles. Which I think we should avoid in favor of a real dependency.

cquiroz:
@golem131 sounds good. It would be nice for this feature to have some documentation too

golem131:
@cquiroz anyway, we need tests for this feature

derekmorr:
I'm hesitant to merge as-is. It seems to add a lot of shapeless magic, and I'm not sure what benefits there is.

darkfrog26:
I agree with @derekmorr, I don't see much benefit of this other than introducing a lot of extra garbage.  Also, I'm generally opposed to adding Shapeless as a dependency to this project if we can avoid it.

garyKeorkunian:
I was hesitant as well.  Which is why that one sat there for so long.  An interesting take, but not sure what real benefit it provides.

darkfrog26:
Then I'll be the bad guy here and close it. :)

PR #141, opened 2016-12-14
Improve scala.js build and test squantsJS on travis

cquiroz:
It seems scala.js tests are not explicitly executed on travis. This PR explicitly calls them it removes some deprecated settings from the scala.js build

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,7 +48,7 @@ before_cache:
 script:
   # Your normal script
   - sbt ++$TRAVIS_SCALA_VERSION test:compile test:fastOptJS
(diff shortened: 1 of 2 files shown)


cquiroz:
Could the new commiters discuss a bit how to proceed from now on. I seem to have rights to commit, but I think PRs be reviewed and commited by somebody else

It would be good to actually make the rules more strict. e.g. requiring at least one reviewer and a passing travis build

darkfrog26:
I agree. How about we agree that a PR shouldn't be merged by the creator and for more complex PRs there is general consensus that the changes are desired.

cquiroz:
@darkfrog26 Sounds good to me. I like when the project settings basically forbid to merge PRs without an accept, that also avoids accidents pressing merge by mistake (Yes, it has happened to me 😄 )

Issue #139, opened 2016-12-13
Seek new maintainers, fork into Typelevel github organization and publish under org.typelevel


milessabin:
Following up on #137, this is a call for new maintainers under the [typelevel github organization](https://github.com/typelevel). Other than @garyKeorkunian the current contributors are listed as,

* Jeremy Apthorp ([nornagon](https://github.com/nornagon))
* Steve Barham ([stevebarham](https://github.com/stevebarham))
* Derek Morr ([derekmorr](https://github.com/derekmorr))
* rmihael ([rmihael](https://github.com/rmihael))
* Florian Nussberger ([fnussber](https://github.com/fnussber))
* Ajay Chandran ([ajaychandran](https://github.com/ajaychandran))
* Gia Bảo ([giabao](https://github.com/giabao))
* Josh Lemer ([joshlemer](https://github.com/joshlemer))
* Dave DeCarpio ([DaveDeCaprio](https://github.com/DaveDeCaprio))
* Carlos Quiroz ([cquiroz](https://github.com/cquiroz))

and it'd be good to see one or more of them step forward. I imagine that everyone who has commented on #137 is also a stakeholder in some sense so they would be very welcome to step forward too.

Given that @garyKeorkunian approached Typelevel for member status I think it's appropriate to continue the project in the Typelevel organization. That being the case I think it would be good if volunteer maintainers shared that commitment.

cquiroz:
Hi, I'd be happy to keep squants under the typelevel umbrella

darkfrog26:
I've already volunteered myself, so I guess I'll do so again here.

vitorsvieira:
Me too!

golem131:
Plus me

milessabin:
@sarahgerweck you were the one who broached the topic of new maintainership ... would you like to volunteer as well?

golem131:
@milessabin what's next?

milessabin:
1. I fork.
2. I add the people volunteering here as contributors.
3. We arrange for one of the above to be able to publish to Sonatype/Central under the "org.typelevel" organization.

On (3) I propose that person be @cquiroz if that's acceptable to everyone here.

sarahgerweck:
@milessabin I haven't worked on the code at all but I've liked this library, so yeah I'm willing to be on the list if we're setting up a pool of maintainers in @garyKeorkunian's absence. I concur that @cquiroz makes sense as the lead maintainer since he's a previous contributor.

[...]

PR #143, opened 2016-12-14
Change organization

cquiroz:
As suggested at #139 we should likely rename the organization to `org.typelevel` and bump the version number. I used `1.0.0-SNAPSHOT` but we could discuss if you prefer to be less agressive

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -5,7 +5,7 @@ import com.typesafe.sbt.osgi.SbtOsgi
 import com.typesafe.sbt.osgi.SbtOsgi.autoImport._
 
 object Versions {
-  val Squants = "0.7.1-SNAPSHOT"
+  val Squants = "1.0.0-SNAPSHOT"
   val Scala = "2.11.8"
   val ScalaCross = Seq("2.12.1", "2.11.8", "2.10.6")
 
@@ -29,7 +29,7 @@ object Resolvers {
 
 object Project {
   val defaultSettings = Seq(
-    organization in ThisBuild := "com.squants",
+    organization in ThisBuild := "org.typelevel",
 
     name := "Squants",
 


golem131:
@cquiroz I think if until tomorrow we will not see any objections, we can do it (24 hour countdown)

cquiroz:
Sounds good 👍

darkfrog26:
I think the code should also be refactored as part of this PR to reference the `org.typelevel` package.

cquiroz:
It seems many typelevel projects use just the name for their packages, at least shapeless, cats and doobie do that. A package change should break old code too

milessabin:
You can use whatever package prefix (or none) that suits you. There's no need to change from "com.squants" unless you want to ... it's orthogonal to the artefact organization you publish under.

larsrh:
Just as a heads-up: You are free to use typelevel.org/squants for the website, or use a custom domain.

darkfrog26:
We're not Java, new versions are allowed to break code. ;)

The reason I want to move to `org.typelevel` is because I don't think we have the desire to maintain and pay for a domain on an ongoing basis.  Perhaps I'm wrong and someone just loves the idea of spending a minimum of $12/year just so we can have a special domain for this?

cquiroz:
@darkfrog26 Do you want to change the domain name of the documentation to `org.typelevel` or the package name? I'm totally ok with the domain name, somewhat I like the plain `squants` package name in use now

darkfrog26:
I lean towards changing the package name.  It also gives us a bit more credibility being tied to TypeLevel, right? :)

milessabin:
FWIW, I don't think that there are any projects which use `org.typelevel` as a package prefix.

darkfrog26:
@milessabin thanks, that's useful.  Then I'll recant my statement and vote to stick with `squants`. :)

PR #162, opened 2016-12-28
Implementation of ceil/floor/rint for Quantities

cquiroz:
Often I need to truncate my Quantities and I need to extract the value, truncate and then rebuild the quantity. This PR would let me do it directly on a `Quantity`

### DIFF
--- a/shared/src/main/scala/squants/Quantity.scala
+++ b/shared/src/main/scala/squants/Quantity.scala
@@ -125,6 +125,24 @@ abstract class Quantity[A <: Quantity[A]] extends Serializable with Ordered[A] {
   def abs: A = unit(math.abs(value))
 
   /**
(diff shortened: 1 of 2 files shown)


cquiroz:
Thanks for your comments. I'll add more documentation as suggested. You're right that unless you know the details of the functions in `scala.math` the differences between `rint/ceil/floor` is subtle and not very well known

I agree that being `Quantity` always a `Double` is a bit odd and limits e.g. `round` which would be a useful method

cquiroz:
About bookkeeping I don't think we have a policy but what you suggest sounds good to me. I guess larger changes will naturally fall into the category of requiring more approvals or discussions while smaller changes don't require so much ceremony

Feel free to create a page on the wiki where these policies could be discussed

PR #173, opened 2017-01-28
Make AngularVelocity a derivative of Angle

shadaj:
This makes it possible to multiply `AngularVelocity` by `Time` to get `Angle`, and divide `Angle` by `Time` to get an `AngularVelocity`.

### DIFF
--- a/shared/src/main/scala/squants/space/Angle.scala
+++ b/shared/src/main/scala/squants/space/Angle.scala
@@ -9,7 +9,8 @@
 package squants.space
 
 import squants._
-import squants.time.Time
+import squants.motion.{AngularVelocity, RadiansPerSecond}
+import squants.time.{Time, TimeIntegral}
 
 /**
  * @author  garyKeorkunian
@@ -18,7 +19,7 @@ import squants.time.Time
  * @param value value in [[squants.space.Radians]]
  */
 final class Angle private (val value: Double, val unit: AngleUnit)
-    extends Quantity[Angle] {
+    extends Quantity[Angle] with TimeIntegral[AngularVelocity] {
 
   def dimension = Angle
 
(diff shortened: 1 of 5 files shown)


derekmorr:
Hi,

Thanks for the PR. I put a few minor comments in-line.

Would you also consider updating the tests? I'm thinking space/AngleSpec.scala. I'm wondering if space/SpaceChecks.scala should also be updated to have a general "AngularVelocity = Angle/Time" property.

-derek

shadaj:
I've added unit tests in AngleSpec, AngularVelocitySpec, and SpaceChecks that verify that transformations between Angle and AngularVelocity work correctly.

I can't see the in-line comments, however. Can you resubmit them to GitHub?

derekmorr:
Thanks for adding the tests, and the quick turn around.

My comments were around using Degrees and DegreesPerSecond in AngularVelocity and Angle, respectively. We try to use SI units when possible, meaning Radians in this case. It looks like you changed AngularVelocity.scala to use Radians in the updated PR, but Angle.scala still uses DegreesPerSecond on line 39. Could you change it to use RadiansPerSecond?

Other than that, this looks good.

-derek

shadaj:
Awesome! I updated Angle.scala to use RadiansPerSecond.

shared/src/main/scala/squants/space/Angle.scala
```diff
@@ -34,6 +35,10 @@ final class Angle private (val value: Double, val unit: AngleUnit)
   def tan = math.tan(toRadians)
   def asin = math.asin(toRadians)
   def acos = math.acos(toRadians)
+
+  protected def timeDerived: AngularVelocity = RadiansPerSecond(toRadians)
```
cquiroz: I wonder why on `AngularVelocity` the protected methods are `[squants]` protected but not in `Angle`

cquiroz:
Thanks for the PR, it looks very good to me

PR #183, opened 2017-02-07
Properly parse temperatures with written in scientific notation.

cranst0n:


cquiroz:
Interesting, do you think this could be applied to other units besides temperature?

cranst0n:
@cquiroz Best I can tell most units use [this](https://github.com/typelevel/squants/blob/90ac0936fcea5be2957b881170140239478a2593/shared/src/test/scala/squants/experimental/Dimension.scala#L73) regex for parsing so they shouldn't share the same issue. That's actually where I pulled the fix from.

cquiroz:
Would you add a comment about this change to the wiki?

PR #187, opened 2017-02-08
Xlint fixes tests

### LINKED ISSUE #185: Xlint warnings
If you enable Xlint, the compiler complains about automatic tuple-ization in the tests:

```
[info] Compiling 89 Scala sources to /home/derek/squants/js/target/scala-2.11/test-classes...
[warn] /home/derek/squants/shared/src/test/scala/squants/QuantitySpec.scala:482: Adapting argument list by creating a 2-tuple: this may not be what you want.
[warn]         signature: BeWord.apply(right: Any): org.scalatest.matchers.Matcher[Any]
[warn]   given arguments: 10.22, "th"
[warn]  after adaptation: BeWord((10.22, "th"): (Double, String))
[warn]     x.toTuple should be(10.22, "th")
[warn]                        ^
[warn] /home/derek/squants/shared/src/test/scala/squants/QuantitySpec.scala:487: Adapting argument list by creating a 2-tuple: this may not be what you want.
[warn]         signature: BeWord.apply(right: Any): org.scalatest.matchers.Matcher[Any]
[warn]   given arguments: 10220, "th"
[warn]  after adaptation: BeWord((10220, "th"): (Int, String))
[warn]     x.toTuple(Thangs) should be(10220, "th")
```

It also complains about declaring an Exception inside a package object:

```
[warn] /home/derek/squants/shared/src/main/scala/squants/market/package.scala:69: it is not recommended to define classes/objects inside of package objects.
[warn] If possible, define class NoSuchExchangeRateException in package market instead.
[warn]   class NoSuchExchangeRateException(val s: String) extends Exception
```

We can't fix this without breaking the public API, but we can fix the warning in the tests. A PR is forthcoming.

derekmorr:
This fixes #185


derekmorr:
Can I get this merged? I'd rather not merge it myself.

PR #190, opened 2017-02-08
don't parse multiple degree symbols and be more flexible in whitespace when parsing temperature

### LINKED ISSUE #189: Temperature Parsing Improvements
The regex used to parse temperatures may benefit from some tweaks. See [here](https://github.com/melrief/pureconfig/pull/111) for some context.

derekmorr:
This would fix #189. I adapted the regex slightly from the current pureconfig-squants version.

This disallows multiple degree symbols. Arguably this was a bug. If this PR is merged, we should add a note to the release notes that this is technically a breaking change (we're less liberal in what inputs we parse).

We should also look at the other string parsers to see if they're similarly affected.

### DIFF
--- a/shared/src/main/scala/squants/thermal/Temperature.scala
+++ b/shared/src/main/scala/squants/thermal/Temperature.scala
@@ -143,19 +143,13 @@ object Temperature extends Dimension[Temperature] with BaseDimension {
   def apply(s: String): Try[Temperature] = {
     val regex = "([-+]?[0-9]*\\.?[0-9]+(?:[eE][-+]?[0-9]+)?)*[ ]*°? *(f|F|c|C|k|K|r|R)".r
     s match {
-      case regex(value, Fahrenheit.symbol) ⇒ Success(Fahrenheit(value.toDouble))
-      case regex(value, "f")               ⇒ Success(Fahrenheit(value.toDouble))
-      case regex(value, "F")               ⇒ Success(Fahrenheit(value.toDouble))
-      case regex(value, Celsius.symbol)    ⇒ Success(Celsius(value.toDouble))
-      case regex(value, "c")               ⇒ Success(Celsius(value.toDouble))
-      case regex(value, "C")               ⇒ Success(Celsius(value.toDouble))
-      case regex(value, Kelvin.symbol)     ⇒ Success(Kelvin(value.toDouble))
-      case regex(value, "k")               ⇒ Success(Kelvin(value.toDouble))
-      case regex(value, "K")               ⇒ Success(Kelvin(value.toDouble))
-      case regex(value, Rankine.symbol)    ⇒ Success(Rankine(value.toDouble))
-      case regex(value, "r")               ⇒ Success(Rankine(value.toDouble))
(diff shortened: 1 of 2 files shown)


derekmorr:
A second commit removes duplicate temperature parsing code (and removed the extraneous degree matches, since the degree symbol isn't captured by the regex).

shared/src/test/scala/squants/thermal/TemperatureSpec.scala
```diff
@@ -55,6 +55,18 @@ class TemperatureSpec extends FlatSpec with Matchers {
     Temperature("ZZ F").failed.get should be(QuantityParseException("Unable to parse Temperature", "ZZ F"))
   }
 
+  they should "be flexible in parsing strings with regard to degree symbol and whitespace" in {
+    Temperature("10.22 f").get should be (Fahrenheit(10.22))
+    Temperature("10.22 °f").get should be (Fahrenheit(10.22))
+    Temperature("10.22  °f").get should be (Fahrenheit(10.22))
```
cquiroz: I guess this is ok though I can imagine somebody arguing that we should only allow one format. How is this for other units? do we allow e.g. `m   /   s`?

derekmorr: We already allow this syntax with the current Temperature parsing regex. I was just adding tests to make sure I didn't break anything while I was changing the regex.

cquiroz: Excellent 👍

shared/src/main/scala/squants/thermal/Temperature.scala
```diff
@@ -141,21 +141,15 @@ object Temperature extends Dimension[Temperature] with BaseDimension {
   def apply[A](n: A, scale: TemperatureScale)(implicit num: Numeric[A]) = new Temperature(num.toDouble(n), scale)
 
   def apply(s: String): Try[Temperature] = {
-    val regex = "([-+]?[0-9]*\\.?[0-9]+(?:[eE][-+]?[0-9]+)?)[ °]*(f|F|c|C|k|K|r|R)".r
+    val regex = "([-+]?[0-9]*\\.?[0-9]+(?:[eE][-+]?[0-9]+)?)*[ ]*°? *(f|F|c|C|k|K|r|R)".r
```
leifwickland: For consistency, I suggest one of the following:
- enclose space in square brackets both times, i.e. `[ ]*°?[ ]*`
- don't space in square brackets, i.e. ` *°? *`
- take it to eleven and allow all the goofy varieties of horizontal whitespace including tab and unicode, i.e. `[\\p{Zs}\\t]*°?[\\p{Zs}\\t]*`

(See http://www.fileformat.info/info/unicode/category/Zs/list.htm for the meaning of `\p{Zs}`)

derekmorr: Ok this is addressed in the latest version.

cquiroz:
Is this ready to be merged?

derekmorr:
Yes.

PR #195, opened 2017-02-09
Only test openjdk 7 and 8 and node version 6

cquiroz:
This PR reduces the test matrix on travis removing openjdk7 and note version 4. This reduces the build time from 1h to 35m


derekmorr:
I've got some concerns with this.

I'm ok removing openjdk7, although I would like to know why it was added in addition to oracle7.

I'm not sure about removing Node 4. It's still supported in LTS until 1 April 2018 according to https://github.com/nodejs/LTS.

I have a few other ideas for how we can speed up CI runs:

I wonder if we can change the Travis config to only rerun the scalaJS tests when needed. For example, right now we run on 2.10 + openjdk7 + node 4 and run squantsJVM/test and squantsJS/test. Then we go to 2.10 + openjdk7 + node 6 and again run both squantsJVM/test and squantsJS/test. I don't think we need to re-compile and re-test squantsJVM/test, since we already know it works on 2.10 + openjdk7. We'd only need to rerun squantsJS/test. Cutting out the duplicate recompiles and tests should speed up CI runs.

Also, if you push a branch to the typelevel/squants project and then submit a PR, Travis runs the CI twice -- once b/c it's a new branch and once for the PR. A quick work-around is to push to a fork, and do the PR from the fork. But I wonder if we can change Travis not to run unless there's a PR.

cquiroz:
Those are valid concerns. I don't really have knowledge of why to select oracle 7 over openjdk 7 or viceversa

About node, node 4 is fairly old and Scala.js in general uses 6 for testing

I don't know how to make the build do what you suggest on only running some test in certain cases

derekmorr:
I've been looking at other typelevel projects' travis configurations. I haven't found one that goes to the same lengths we go in CI. Most of them just use oracle8 and whatever node version is the default. So I've changed my mind, and this PR looks good to me. Even cutting openjdk7 and node4 we'll still be doing more tests than our peers. :)

cquiroz:
Yes, our travis config seems overkill. OTOH I still think 35m is a lot

derekmorr:
I was reviewing the history and it looks like @golem131 added in the node4 and 6 tests in December 2016 in https://github.com/typelevel/squants/commit/2be0aa00afa3ac016a2bed6a6b21fc0c5b327c5f . Could he comment about why?

derekmorr:
Yeah, 35 min is really long, especially since it will run twice if you push to a feature branch and then do a PR.

derekmorr:
I'll wait to hear back from @golem131 about this. If I don't hear back in a day or so, I'll approve the PR so it can get merged.

derekmorr:
Haven't heard anything back from @golem131 yet, so I'll approve and merge.

PR #200, opened 2017-02-11
parse Tuples with any numeric anyval.

### LINKED ISSUE #198: Parsing tuples only works if the value is a double
According to the README, this should work:

```scala
val power = Power((40, "MW"))
```

but it doesn't:

```scala
% sbt squantsJVM/console
<snip>
Welcome to Scala 2.11.8 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_121).
Type in expressions for evaluation. Or try :help.

scala> val power = Power((40, "MW"))
scala.MatchError: (40,MW) (of class scala.Tuple2)
  at squants.Dimension$class.parse(Dimension.scala:58)
  at squants.energy.Power$.parse(Power.scala:60)
  at squants.energy.Power$$anonfun$apply$1.apply(Power.scala:63)
  at squants.energy.Power$$anonfun$apply$1.apply(Power.scala:63)
  ... 42 elided
```
`Dimension.parse()` only works on `Tuple2[Double, String]` : https://github.com/typelevel/squants/blob/master/shared/src/main/scala/squants/Dimension.scala#L58

derekmorr:
This fixes #198 . 

It's not my ideal solution, but it keeps the public API the same.

### DIFF
--- a/shared/src/main/scala/squants/Dimension.scala
+++ b/shared/src/main/scala/squants/Dimension.scala
@@ -56,9 +56,15 @@ trait Dimension[A <: Quantity[A]] {
    * @return Try[A]
    */
   protected def parse(value: Any) = value match {
-    case s: String              ⇒ parseString(s)
-    case (v: Double, u: String) ⇒ parseTuple(v, u)
+    case s: String              => parseString(s)
+    case (v: Byte, u: String)   => parseTuple(v.toDouble, u)
+    case (v: Short, u: String)  => parseTuple(v.toDouble, u)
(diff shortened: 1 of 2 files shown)


cquiroz:
The code looks fine, I'm still not sure about the use case of doing dimensions with tuples but I guess it is a nice facility if you need it

derekmorr:
Well, I think we have to fix it. The functionality is there, so if we remove it we break backwards-compat. And it doesn't make sense to leave it broken.

cquiroz:
That is very true, I'll approve it

PR #203, opened 2017-02-17
use tut to build docs

### LINKED ISSUE #177: integrate tut to check code samples in docs
We should integrate the tut tool to check that the code samples in the docs compile.

I've pushed a `tut` branch with a very rough first pass at this. Not all of the errors are removed (some of the operator overloading still causes errors b/c of conflicting imports), and I had to make some changes to how the OSGI settings are pulled in, which I haven't tested yet. The `tut` branch right is just to give folks an idea of how we might integrate tut into the docs (it's absolutely not ready to be merged yet).

derekmorr:
This resolves #177 .

Most of the bugs I've submitted over the past few weeks have been found by running tut on the README. All of the examples in the README now compile with tut.

I made some stylistic changes in the README in this PR. I made the imports more prominent, since I assumed users could get confused about what imports they need (I know I was when i was new to the project). If reviews don't like that change, I can remove it and just pull in everything like the `squantsJVM/console` task does.

If this PR gets merged, I'll submit a followup PR to run tut as part of the Travis checks.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -12,7 +12,10 @@ lazy val squants = crossProject
   .in(file("."))
   .settings(defaultSettings: _*)
   .jvmSettings(
-    osgiSettings: _*
+    osgiSettings,
+    tutSettings,
+    tutTargetDirectory := file("."),
+    tutSourceDirectory := file("shared") / "src" / "main" / "tut"
   )
   .jsSettings(
     parallelExecution in Test := false,
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -7,3 +7,5 @@ addSbtPlugin("com.typesafe.sbt" % "sbt-osgi" % "0.8.0")
 addSbtPlugin("com.jsuereth" % "sbt-pgp" % "1.0.0")
 
 addSbtPlugin("org.xerial.sbt" % "sbt-sonatype" % "1.1")
+
+addSbtPlugin("org.tpolecat" % "tut-plugin" % "0.4.8")
--- a/shared/src/main/tut/README.md
+++ b/shared/src/main/tut/README.md
@@ -78,15 +78,10 @@ or produced, that is, Power is the first time derivative of Energy.
 
 Consider the following code:
 
-```scala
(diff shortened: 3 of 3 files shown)


README.md
```diff
@@ -97,63 +98,115 @@ Because this code compiles, detection of these errors is pushed further into the
...
-automatically applying scale and type conversions at run-time.  For example,
+automatically applying scale and type conversions at run-time.  For example:
 
 ```scala
-val load1: Power = Kilowatts(12)        // returns Power(12, Kilowatts) or 12 kW
-val load2: Power = Megawatts(0.023)     // Power: 0.023 MW
-val sum = load1 + load2                 // Power: 35 kW - unit on left side is preserved
-sum should be(Kilowatts(35))            
-sum should be(Megawatts(0.035))         // comparisons automatically convert scale
+scala> import squants.energy.{Kilowatts, Megawatts, Power}
```
cquiroz: Is this the output of `tut`?

derekmorr: The imports? No, I added that.

derekmorr:
Good question. Yes, the official readme is now the one in `shared/src/main/tut`. When you run `sbt tut` it checks all of the code examples in `shared/src/main/tut/README.md` and produces the README in the root directory.

If ther eare code examples you want checked by tut you annotate then with 

    ```tut
    //  code
    ```

instead of

    ```scala
    //  code
    ```
You can still use code samples without the "tut" annotation -- tut will ignore them.

cquiroz:
That's great, Hopefully we could later on build the public documentation in `squants.org` out of this as well
Do you think we should at this moment make release 1.2.0?

derekmorr:
Thanks.

Yes, I think 1.2.0 is a good idea. We have plenty of queued up changes. The tut docs were the last thing I was hoping to get into 1.2.0. 

Since you approved this PR, I'll send in one more to run `sbt tut` as part of the Travis builds. Then I think we're good for 1.2 unless any other maintainers have objections.

cquiroz:
Excellent, let me know and I'll make a release

PR #209, opened 2017-03-05
Added Solar units of measure

robotsnowfall:
| Measure | Unit | Reference |
| ------- | ---- | --------- |
| Length | (Nominal) Solar Radius | https://en.wikipedia.org/wiki/Solar_radius |
| Power | Solar Luminosity | https://en.wikipedia.org/wiki/Solar_luminosity |
| Mass | Solar Mass | https://en.wikipedia.org/wiki/Solar_mass |

First time submission, let me know if I missed something.

### DIFF
--- a/shared/src/main/scala/squants/energy/Power.scala
+++ b/shared/src/main/scala/squants/energy/Power.scala
@@ -52,6 +52,7 @@ final class Power private (val value: Double, val unit: PowerUnit)
   def toGigawatts = to(Gigawatts)
   def toBtusPerHour = to(BtusPerHour)
   def toErgsPerSecond = to(ErgsPerSecond)
+  def toSolarLuminosities = to(SolarLuminosities)
 }
 
 /**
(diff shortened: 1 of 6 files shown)


derekmorr:
Thanks for the patch. The changes look good (I had to double check that some of the changes didn't belong in the radio package, but I think they're good where they are).

A few questions:

* Should nominal solar radius be added as well?
* Are there alternate symbols that don't involve unicode symbols? In the past, some users have raised objections about certain symbols being difficult to enter on a keyboard.

robotsnowfall:
I think having nominal solar radius would be useful, yes. I have a patch that adds it.

As for the unicode symbols, I totally see where you're coming from. I saw the symbol for Angstrom (`Å`) was already in the repo, so I didn't think it would be too much of a stretch.  There doesn't seem to be generally accepted symbols other than the ones provided. I suppose we could make them up though? There would be a danger of overlap if we chose something like `SR` or `RS` for solar radius, or `SL`/`LS` for solar luminosity, etc.

cquiroz:
Thanks, it looks good to me.
I'm ok with unicode units if they are in common use or officially accepted for the unit

derekmorr:
Ok, the patch looks good to me. 

I'd rather not invent new symbols. If there are only unicode symbols, then we'll use them.

Thanks again for the PR!

PR #210, opened 2017-03-16
Pressure units

cquiroz:
Added a couple of units of pressure used often in medicine and engineering

* mmHg
* Torr

### DIFF
--- a/shared/src/main/scala/squants/motion/Pressure.scala
+++ b/shared/src/main/scala/squants/motion/Pressure.scala
@@ -30,11 +30,12 @@ final class Pressure private (val value: Double, val unit: PressureUnit)
   def *(that: Area): Force = Newtons(toPascals * that.toSquareMeters)
   def *(that: Time) = ??? // returns DynamicViscosity
 
-  def toPascals = to(Pascals)
-  def toBars = to(Bars)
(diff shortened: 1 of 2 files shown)


shared/src/main/scala/squants/motion/Pressure.scala
```diff
@@ -68,18 +70,32 @@ object StandardAtmospheres extends PressureUnit {
...
 }
 
+object MillimeterOfMercury extends PressureUnit {
+  val symbol = "mmHg"
+  val conversionFactor = Newtons.conversionFactor * 133.322387415
+}
+
+object Torr extends PressureUnit {
+  val symbol = "Torr"
+  val conversionFactor = Newtons.conversionFactor * 101325d / 760d
```
derekmorr: A Torr is "defined as exactly 1 / 760 of a standard atmosphere" (per wikipedia). So I wonder if `conversionFactor` would be better defined as `StandardAtmospheres.conversionFactor / 760d` ? It won't matter for correctness, but might make the definition easier to read.

cquiroz:
Did as suggested and also changed `MillimeterOfMercury` and `Torr` class names to the plural version to match the other units

PR #213, opened 2017-03-30
Frequency Enhancements

### LINKED ISSUE #134: multiplication with TimeDerivative is not supported
Would it be possible to support multiplication with a TimeDerivative like Frequency ?

here is an example: 

```
    val averageMessageSize: Information = Bytes(600)
    println(averageMessageSize)

    val messageRate: Frequency = Each(7000) / Seconds(1)
    println(messageRate)

    // A * Frequency is not supported
    // following doesn't compile:
    //    val averageDataRate: DataRate = averageMessageSize * messageRate
    // But A / Time is supported, so convert Frequency to Time and divide
    val averageDataRate: DataRate = averageMessageSize / (Each(1) / messageRate)
    println(averageDataRate in MebibytesPerSecond)

```

garyKeorkunian:
Added /(Time) and per(Time) to Squantified numbers to support expression like:

```scala
val f: Frequency = 1000 / second
val f: Frequency = 1000 per second
```

Added *(Frequency) to TimeIntegral to support conversions like:

```scala
val dr: DataRate = 1000.bytes * 10/second
```

Added *(TimeIntegral) to Frequency to support the same conversion with reversed operands:

```scala
val dr: DataRate = 10/second * 1000.bytes
```

However, this last expression returns type Any and needs to be reviewed for ideas on how to fix that before this is merged.

This is intended to close #134

### DIFF
--- a/README.md
+++ b/README.md
@@ -802,14 +802,20 @@ scala> val low = 40.dollars / megawattHour
 low: squants.market.Price[squants.energy.Energy] = 40.0 USD/1.0 MWh
 ```
 
-Implicit conversion support for using Double on the left side of multiplication:
+Implicit conversion support for using Doubles, Longs and BigDecimals on the left side of multiply and divide operations:
 
 ```scala
-scala> val load = 10 * 4.MW
-load: squants.energy.Power = 40.0 MW
+scala> val load = 10.22 * 4.MW
+load: squants.energy.Power = 40.88 MW
 
 scala> val driveArrayCapacity = 12 * 600.gb
 driveArrayCapacity: squants.information.Information = 7200.0 GB
+
+scala> val freq = 60 / second
+freq: squants.time.Frequency = 60.0 Hz
+
+scala> val freq2 = BigDecimal(36000000) / hour
+freq2: squants.time.Frequency = 10000.0 Hz
 ```
 
 Create Quantity Ranges using `to` or `plusOrMinus` (`+-`) operators:
--- a/shared/src/main/scala/squants/time/Frequency.scala
+++ b/shared/src/main/scala/squants/time/Frequency.scala
@@ -9,6 +9,10 @@
 package squants.time
 
 import squants._
+import squants.electro.{ ElectricCharge, ElectricPotential, MagneticFlux }
+import squants.information.{ DataRate, Information }
+import squants.motion.{ AngularVelocity, Pressure, PressureChange, Yank }
+import squants.photo.{ Illuminance, LuminousEnergy, LuminousExposure, LuminousFlux }
 
 /**
  * Represents a quantity of frequency, which is the number cycles (count) over time
@@ -25,6 +29,24 @@ final class Frequency private (val value: Double, val unit: FrequencyUnit)
 
   protected[squants] def timeIntegrated = Each(toHertz)
   protected[squants] def time = Seconds(1)
+
+  def *(that: Acceleration): Jerk = that * this
+  def *(that: Angle): AngularVelocity = that * this
+  def *(that: Dimensionless): Frequency = this * that.toEach
+  def *(that: ElectricCharge): ElectricCurrent = that * this
+  def *(that: Energy): Power = that * this
+  def *(that: Force): Yank = that * this
+  def *(that: Information): DataRate = that * this
+  def *(that: Length): Velocity = that * this
+  def *(that: LuminousEnergy): LuminousFlux = that * this
+  def *(that: LuminousExposure): Illuminance = that * this
+  def *(that: MagneticFlux): ElectricPotential = that * this
+  def *(that: Mass): MassFlow = that * this
+  def *(that: Momentum): Force = that * this
+  def *(that: Power): PowerRamp = that * this
+  def *(that: Pressure): PressureChange = that * this
+  def *(that: Velocity): Acceleration = that * this
+  def *(that: Volume): VolumeFlow = that * this
 
   def toHertz = to(Hertz)
   def toKilohertz = to(Kilohertz)
--- a/shared/src/test/scala/squants/QuantitySpec.scala
+++ b/shared/src/test/scala/squants/QuantitySpec.scala
@@ -8,12 +8,11 @@
 
 package squants
 
-import org.scalatest.{FlatSpec, Matchers}
-import squants.thermal.{Celsius, Fahrenheit}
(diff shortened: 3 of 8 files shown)


garyKeorkunian:
It may be that the Frequency.*(TimeIntegral) method is too abstract and I'll need to define a method for each TimeIntegral in the libary ....

```scala
class Frequency(...) {
   ...
   def *(that: Length): Velocity = that * this
   def *(that: Velocity): Acceleration = that * this
   def *(that: Information): DataRate = that * this
   // etc ...
}
```

It would be nice if the TimeIntegral/Derivative abstraction could handle that, but maybe not.

Thoughts?

garyKeorkunian:
I replaced the Frequency * TimeIntegral code, which was limited, with explicit methods for each defined TimeIntegral.  This achieves the API requirement, but with a bit more code.

This could be improved in the future, but this PR should be mergeable as is - pending review.

shared/src/main/scala/squants/time/Frequency.scala
```diff
@@ -26,6 +30,36 @@ final class Frequency private (val value: Double, val unit: FrequencyUnit)
   protected[squants] def timeIntegrated = Each(toHertz)
   protected[squants] def time = Seconds(1)
 
+  // Below are two failed attempts to invert the TimeIntegral * Frequency method to support Frequency * TimeIntegral
```
derekmorr: I think we should remove the comments from lines 33 - 44 into the PR. I'd like to preserve them, but I don't think they need to be in the source.

shared/src/test/scala/squants/QuantitySpec.scala
```diff
@@ -10,7 +10,7 @@ package squants
 
 import org.scalatest.{ FlatSpec, Matchers }
 import squants.thermal.{ Celsius, Fahrenheit }
-import squants.time.Hours
+import squants.time.{ Hours, Hertz }
```
derekmorr: nit - Hertz should be before Hours to keep imports alphabetized.

derekmorr:
I agree spelling out the specific types is sub-optimal, but at least we're type safe. We might be able to implement it more generically in the future at the cost of binary compatibility.

With this new syntax, we should update the README. I had removed this syntax in #202 a few months ago. Can you update `shared/src/main/tut/README.me` and readd 

```scala
val freq = 60 / second
```

example in the "Implicit conversion support for using Double on the left side of multiplication" section (which we probably should rename now).

(this was the original line https://github.com/typelevel/squants/pull/202/files#diff-04c6e90faac2675aa89e2176d2eec7d8L487)

shared/src/test/scala/squants/time/FrequencySpec.scala
```diff
@@ -62,6 +69,26 @@ class FrequencySpec extends FlatSpec with Matchers {
...
+    Hertz(100) * WattHours(100) should be(Watts(10000))
+    Hertz(100) * Newtons(100) should be(NewtonsPerSecond(10000))
+    Hertz(100) * Bytes(100) should be(BytesPerSecond(10000))
+    Hertz(100) * Meters(100) should be(MetersPerSecond(10000))
+    Hertz(100) * LumenSeconds(100) should be(Lumens(10000))
+    Hertz(100) * LuxSeconds(100) should be(Lux(10000))
+    Hertz(100) * Webers(100) should be(Volts(10000))
+    Hertz(100) * Kilograms(100) should be(KilogramsPerSecond(10000))
+    Hertz(100) * NewtonSeconds(100) should be(Newtons(10000))
+    Hertz(100) * Watts(100) should be(WattsPerHour(10000))
```
derekmorr: Why is this in `WattsPerHour`? Shouldn't the unit be watts per second?

garyKeorkunian: Good catch.  Most time derivative relationships are based on the second.  Energy/Power/PowerRamp prefer the WattHour, Watt and WattsPerHour units, so the base time is one hour.  The new conversion assumed 1 second, which coupled with my blind paste and copy test setup, led to a passing test on a bad conversion.  It's fixed up now.

Thanks!

garyKeorkunian: Although WattsPerSecond, which is effectively a joule / sec^2

README.md
```diff
@@ -788,14 +789,21 @@ scala> val low = 40.dollars / megawattHour
 low: squants.market.Price[squants.energy.Energy] = 40.00 USD/1.0 MWh
 ```
 
-Implicit conversion support for using Double on the left side of multiplication:
+Implicit conversion support for using Doubles, Longs and BigDecimals on the left side of multiply and divide operations:
```
derekmorr: It looks like you edited `README.md` directly. The readme is built via tut now, so these changes will be overwritten on the next build. Can you also update `shared/src/main/tut/README.md` so they're preserved?

garyKeorkunian: Ah, yes.  Thanks.  Fixed.

garyKeorkunian: One thing I noticed is 

`BigDecimal(1) / hour` works in the tut, but
`BigDecimal(1) per hour` does not:  method not defined.

However, both of these are defined in the same implicit class and both work in the console.  For the README, I just went back to using the / operator.  Any thoughts on that?

shared/src/main/scala/squants/time/Frequency.scala
```diff
@@ -26,6 +30,24 @@ final class Frequency private (val value: Double, val unit: FrequencyUnit)
...
+  def *(that: Dimensionless): Frequency = this * that.toEach
+  def *(that: ElectricCharge): ElectricCurrent = that * this
+  def *(that: Energy): Power = that * this
+  def *(that: Force): Yank = that * this
+  def *(that: Information): DataRate = that * this
+  def *(that: Length): Velocity = that * this
+  def *(that: LuminousEnergy): LuminousFlux = that * this
+  def *(that: LuminousExposure): Illuminance = that * this
+  def *(that: MagneticFlux): ElectricPotential = that * this
+  def *(that: Mass): MassFlow = that * this
```
cquiroz: Will this produce e.g. [kgs/sec]

garyKeorkunian: Yes, it will.

[T^1] * [M] = [M/T]

derekmorr:
LGTM.

Issue #212, opened 2017-03-27
Best Unit For


darkfrog26:
I've been using a little utility class I wrote that I think would make a lot of sense if the concept were integrated on global level in Squants:

```scala
  private val conversions = List(Bytes, Kilobytes, Megabytes, Gigabytes, Terabytes, Petabytes, Exabytes, Zettabytes, Yottabytes).reverse

  def bestUnitFor(i: Information, minimumUnit: InformationUnit = Bytes): Information = {
    val unit = conversions.find { u =>
      u == minimumUnit || i.to(u) > 1.0
    }.get

    val value = i.to(unit)
    unit(value)
  }
```

This is explicitly defined currently for `Information` (file sizes) but could be implemented for everything. The biggest hurdle for this would be the necessity of having an ordered representation of levels (see `conversions` above).

darkfrog26:
To clarify the use-case I use it to determine the "best unit" before I format the value for output:

```scala
  def format(i: Information): String = {
    f"${i.value}%2.2f ${i.unit.symbol}"
  }
```

derekmorr:
I'm not sure the best way to implement this. For example, here you're using the base-10 sizes, but a user might also want base-2 (Kibibytes, Mebibytes, etc). The same would apply for imperial vs metric units, for example.

I like the idea, I'm just not sure how to integrate it without requiring the caller to specify a list of conversions. If we had a better concept of groups of units in squants (CGS, metric, etc), it might be easier to do this.

darkfrog26:
I agree, I was tossing this out more for feedback than anything.  If it were trivial I would have just submitted a PR. ;)

garyKeorkunian:
How about something like this:

```scala
def inBestUnit(min: Double = 1.0, max: Double = 1000.0,
                 units: Set[UnitOfMeasure[A]] = dimension.units,
                 default: UnitOfMeasure[A] = dimension.siUnit) = {
    units.find { u =>
      to(u) >= min && to(u) <= max
    } match {
      case Some(u) => in(u)
      case None => in(default)
    }
  }
```

This would be a method of the Quantity class.  It provides a set of overridable defaults to make basic usage simple, yet can be customized with a specific range, limited set of units, and a default if no unit in the `units` results in a value within the range.

Basic usage is limited by the fact that the `dimensions.units` list is not strictly ordered (although we could sort on the conversion factor first).  The overrides may seem clunky in that you have to supply the full ordered list you want to choose from, but it's not too bad.

[...]

PR #215, opened 2017-04-03
First draft of best-fit

derekmorr:
This is a first draft of defining groups of `UnitOfMeasure` and formatters with best fit. See #212 for motivation. This PR is not ready to be merged, and is here for discussion only for the time being.

I defined some new traits:

- `squants.units.Units` defines a group of related `UnitOfMeasure[A]`'s (for example, Inches, Feet, Yards, Miles). It uses a new `UomOrdering` to provide a lazy val with a `SortedSet` of the units.
- `squants.units.SIUnits` has some sample `Units` implementations for various SI units. It's incomplete right now; it's provided as an example of how to use the `Units` trait. There are other examples in the `Units` package object for some non-SI units.

- `squants.formatter.Formatter` defines the `bestUnit()` function for picking the best unit for a quantity. Implementation is entirely user-defined.
- `squants.formatter.DefaultFormatter` is a default implementation of `Formatter` which uses a `SortedSet[UnitOfMeasure[A]]` to pick a value. This will probably work well-enough for most use-cases, but is broken out into its own trait to keep it separate from `Formatter`.
- `squants.formatter.Formatters` has some sample `Formatter` implementations using some of the `Units` defined above.

For all of this, the names and organization of the new traits should be discussed. In particular, I'm not sure `Units` is the best name for a group of related `UnitOfMeasure`. Also, in `squants.units.SIUnits` I'm not sure if we want mirror the hierarchy in the rest of the project (electro, energy, mass, motion, etc) or just lump things into a single flat namespace as I did now. 

Here's an example of how you can use the new traits and objects to provide the best-fit functionality:

```scala
scala> import squants.formatter._
import squants.formatter._

scala> import squants.units.SIUnits._
import squants.units.SIUnits._

scala> val siLengthFormatter = new DefaultFormatter[Length] { val units = SILengths }
siLengthFormatter: squants.formatter.DefaultFormatter[squants.space.Length]{val units: squants.units.SIUnits.SILengths.type} = $$$c1c91dda193eef78668663dd30698de$$$$w$$anon$1@24bca3c6

scala> List(500.cm, 3000.meters, 120.mm).map(siLengthFormatter.bestUnit)
res0: List[squants.space.Length] = List(5.0 m, 3.0 km, 12.0 cm)
```

(I suspect that an SI-Length formatter should be provided out-of-the-box. This is just an example.)

I've tried to keep functionality separate. `Units` is only for groups of `UnitOfMeasure`. The `Formatter` trait builds on that functionality, but could be just one such consumer of `Units`. Users can provide their own `Unit` implementations or their own `Formatter` implementations without submitting PRs or rebuilding the library.

One downside of this design is that maintainers have to add new `UnitOfMeasure`'s in two places -- in the relevant `Quantity` file and in one or more `Units` collections. For example, to add Furlongs, a PR would have to add it to `shared/src/main/scala/space/Length.scala` and to `shared/src/main/scala/units/ImperialUnits.scala` (once we define `ImperialUnits.scala`). Ideally, I'd like a more automatic way of populating `Units` but I don't know of a good way right now.

Please provide feedback on the API structure, package layout, etc. I'd like to get the basic structure ironed out and committed. Then we can begin the time-consuming task of defining `Units` for all the


garyKeorkunian:
My take ...

* The general approach is sound:  one package for describing the unit sets and another for the formatting concern.
* The name `Units` does seem a bit general for this case.  It could cause confusion between this concept and the entire list of `units` for a Dimension.  Perhaps we call them `UnitGroups`.
* I agree the downside is the need to add new units to multiple places.  Perhaps we can use traits to tag each unit with the group(s) it belongs to.  Then create the groups (lazily) by filtering on the master list.

```scala
trait SiUnit               // currently exists in UnitOfMeasure.scala
trait SiBaseUnit extends SiUnit     // currently exists in UnitOfMeasure.scala
trait ImperialUnit

...

object Meters extends LengthUnit with SiBaseUnit { ... }
object Kilometers extends LengthUnit with SiUnit { ... }
object Inches extends LengthUnit with ImperialUnit { ... }
object Feet extends LengthUnit with ImperialUnit { ... }

...

lazy val SiLengths = Length.units.flatMap {
  case u: SiUnit => Some(u)
  case _ => None
}

lazy val ImperialLengths = Length.units.flatMap {
  case u: ImperialUnit => Some(u)
  case _ => None
}
```

Perhaps using macros to generate the sets is another possibility.  I expect we would still need the trait tags or something like it.

garyKeorkunian:
Here's a more refined way to filter those:

```scala
lazy val SiLengths = Length.units.collect { case u: SiUnit => u }
```

sarahgerweck:
> Perhaps using macros to generate the sets is another possibility. I expect we would still need the trait tags or something like it.
If somebody wants to do this, I have some [code](https://github.com/sarahgerweck/scala-utils/blob/master/macro/src/main/scala-2.11/org/gerweck/scala/util/TypeUtilMacros.scala#L21) that could be adapted to do the work of finding all the known objects that inherit from a trait. (I wouldn't propose to add a dependency on that library: it could be copy-pasted and tweaked.) This part of the macro system was [historically somewhat unreliable](https://issues.scala-lang.org/browse/SI-7046), but that has apparently recently been fixed.

derekmorr:
Gary,

I think the trait approach will work, at least for SI units. Non-SI gets harders, because there's the british version of the imperial system, the US Customary system, etc. I'm not sure we want to define marker traits for all of those, but maybe for a subset it's worth it.

derekmorr:
I should have updated PR this weekend.

derekmorr:
I've renamed `Units` to `UnitGroup`. I added the `SiUnit` marker trait to several SI Units in Length. I also added basic unit tests for `DefaultFormatter`.

If this design meets w/ folks' approval, I'll work through the remanining units to add `SiUnit` and
update the `UnitGroup` definition -- I'm just waiting to do that tedious work until we have the core
design ironed out.

I do have one concern about using the `SiUnit` trait in "core" to define `UnitGroup`s. In the current
design, "core" is only concerned with defining `UnitOfMeasure`s and conversions between them. Now we're 
having it decide what is and what's not SI. That's not quite to cut-and-dried. [The SI allows several
non-SI units to be used in SI](http://www.bipm.org/en/publications/si-brochure/table6.html), for example tonne, litre, hectare, and au). What do we do with these units? Do we decide to make the "SI" and attach the `SiUnit` trait to them? Do we leave them non-SI? It seems that there are valid arguments for both a "strict SI" argument and a "loose SI" argument.

Apologies if this sounds like nit-picking, but once we make a decision about these units, we can't change
it without breaking compatibility.

I propose we only use the `SiUnit` trait on units in the SI. We can create `UnitGroup`s with the additional units if desired.

```scala
object SILengths extends UnitGroup[Length] { 
    val units: Set[UnitOfMeasure[Length]] = Length.units.collect { case si: SiUnit => si }
}

object LooseSILengths extends UnitGroup[Length] { 
    val units: Set[UnitOfMeasure[Length]] = SILengths.units + AstronomicalUnits
}

**Comment by derekmorr** [comment]
FYI, I haven't abandoned this. Just been busy w/ work and haven't had time to get back to it. I have been working on it this weekend. I'm wondering if a structure like this would work:

```scala
object Implicits {
  implicit def implicitFormatter[A <: Quantity[A]](implicit unitGroup: UnitGroup[A]): Formatter[A] = {
    new DefaultFormatter[A] { val units: UnitGroup[A] = unitGroup }
  }

  implicit class EnhancedQuantity[A <: Quantity[A]](quantity: Quantity[A]) {
    def inBestUnit(implicit formatter: Formatter[A]): Quantity[A] = {
      formatter.bestUnit(quantity)
    }
  }
}
```

That should let you do something like:

```scala
scala> implicit object SiLengths extends UnitGroup[Length] {
     |     val units: Set[UnitOfMeasure[Length]] = Length.units.collect { case si: SiUnit => si }
     |   }
defined object SiLengths

scala> val data = List(500.cm, 3000.meters, 120.mm)
data: List[squants.space.Length] = List(500.0 cm, 3000.0 m, 120.0 mm)

scala> import squants.formatter.Implicits._
import squants.formatter.Implicits._

scala> data map { _.inBestUnit }
res2: List[squants.Quantity[squants.space.Length]] = List(5.0 m, 3.0 km, 1.2 dm)
```

That gives us the syntax some folks were discussing earlier, while still keeping the `inBestUnit()` functionality out of `Quantity` itself.

One issue I ran into is for imperial volume units. Because the imperial systems uses different units for dry vs wet volume, I think we'd run into conflicting implicits:

```scala
implicit object UsCustomaryLiquidVolumes extends UnitGroup[Volume] {
    val units: Set[UnitOfMeasure[Volume]] = Set(Teaspoons, Tablespoons, FluidOunces, UsCups, UsPints, UsQuarts, UsGallons)
}

implicit object UsCustomaryDrylVolumes extends UnitGroup[Volume] {
    val units: Set[UnitOfMeasure[Volume]] = Set(UsDryPints, UsDryCups, UsDryQuarts, UsDryGallons)
}
```

derekmorr:
I've restructed the SI units and started defining the non-SI units. I'm trying to use implicits to cut down on boilerplate (sadly I couldn't find a way to do this for the SI unit groups w/o redefining the `UnitGroup` trait, so there's a lot of boilerplate). I'll squash most of the commits and push up some examples and imperial groups.

shared/src/main/scala/squants/formatter/DefaultFormatter.scala
```diff
@@ -0,0 +1,19 @@
...
+
+trait DefaultFormatter[A <: Quantity[A]] extends Formatter[A] {
+  def units: UnitGroup[A]
+
+  override def bestUnit(quantity: Quantity[A]): A = {
+    val unit =
+      units.sortedUnits
+        .takeWhile { u => quantity.to(u).abs >= 1.0 }
+        .lastOption
+        .getOrElse(units.sortedUnits.head)
```
cquiroz: What happens if `units` is empty? `head` would throw an exception

shared/src/main/scala/squants/unitgroups/si/strict/package.scala
```diff
@@ -0,0 +1,254 @@
...
+      val units: Set[UnitOfMeasure[Frequency]] = Frequency.units.collect { case si: SiUnit => si }
+    }
+
+    object SiTimes extends UnitGroup[Time] {
+      val units: Set[UnitOfMeasure[Time]] = Time.units.collect { case si: SiUnit => si }
+    }
+  }
+
+
+  import electro._
```
cquiroz: Why do you need the imports here?

cquiroz:
This looks quite nice. I guess it could benefit from some documentation though.
A cleanup of the history would be good too, there are a bunch of merge commits

derekmorr:
I managed to get the implicits fixed, and I'm going to push those as multiple smaller PRs.

shared/src/main/scala/squants/unitgroups/si/strict/package.scala
derekmorr: These are all going to go away in the next push.

garyKeorkunian:
Is it a good idea to merge a PR with 39 commits?  Should they be squashed?

derekmorr:
At this point, the branch has so many changes, I'm going to break this PR up into small ones.

derekmorr:
I pushed up #234 and #235 as the first part of breaking this PR up into smaller pieces.

Once those are approved I'll push up the formatters.

PR #217, opened 2017-04-06
Money Improvements and Rounding

### LINKED ISSUE #216: Money rounding mode
Looking at the code I've noticed that `Money.toString` uses the explicit rounding mode `HALF_EVEN`, while `Money.toFormattedString` does not. Is this the intended behaviour? 

Speaking about rounding modes, I'd find useful a method to apply rounding to a `Money` instance. What I'm doing in my code right now is something like this:

```scala
implicit class MoneyOps(m: Money) {
  def rounded: Money = Money(
    value = m.amount.setScale(m.currency.formatDecimals, BigDecimal.RoundingMode.HALF_EVEN),
    currency = m.currency
  )
}
```

This may be useful to apply rounding in calculations on purpose.

garyKeorkunian:
* added `Quantity.rounded`
* added `Money.rounded`
* modified `Money.toString` to use unrounded value
* fixed `Money.toFormattedString` to use proper rounding
* added `Money.mapAmount`, which is applied to the `amount` field
* fix README

Fixes #216

### DIFF
--- a/README.md
+++ b/README.md
@@ -346,6 +346,8 @@ newLoad: squants.energy.Power = 14.0 kW
 
 The `q.map(f)` method effectively expands to `q.unit(f(q.to(q.unit))`
 
+NOTE - For Money objects, use the `mapAmount` method as this will retain the BigDecimal precision used there.
+
 ### Approximations
 Create an implicit Quantity value to be used as a tolerance in approximations.
 Then use the `approx` method (or `=~`, `~=`, `≈` operators) like you would use the `equals` method (`==` operator).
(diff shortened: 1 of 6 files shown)


garyKeorkunian:
Please note, while this should not break binary compatibility, it does change the behavior of Money.toString and Money.toFormattedString.

README.md
```diff
@@ -621,25 +635,25 @@ scala> val exchangeRates = List(USD / CAD(1.05), USD / MXN(12.50), USD / JPY(100
...
+moneyContext: squants.market.MoneyContext = MoneyContext(squants.market.USD$@dfad661,Set(squants.market.GBP$@69f4ad73, squants.market.BRL$@7d0734cf, squants.market.CNY$@bce5be3, squants.market.USD$@dfad661, squants.market.HKD$@5959188e, squants.market.DKK$@259bd19, squants.market.RUB$@11b0980e, squants.market.BTC$@79056cc3, squants.market.ARS$@725ae543, squants.market.SEK$@6562b9fe, squants.market.XAG$@431dce40, squants.market.CZK$@360576, squants.market.CAD$@2695bcd0, squants.market.MXN$@3cfd1c9a, squants.market.KRW$@2df1ecc2, squants.market.NZD$@45fd55e7, squants.market.CHF$@3c035bc, squants.market.CLP$@4fabbf95, squants.market.NOK$@67c83220, squants.market.JPY$@61224aa0, squants.market.XAU$@6c60e9a9, squants.market.MYR$@14cc0d53, squants.market.EUR$@101cfdd0, squants.market.INR$@1627...
```
derekmorr: Not pertinent to this PR, but it would be nice to have stable toString()'s for MoneyContext so this output didn't change so often).

derekmorr:
I'm concerned we may get some bugs filed over the `toString()` change (for example #145 was about this).

README.md
garyKeorkunian: I added a toString method to Currency that will output just the code.  That should stabilize the README output.

garyKeorkunian:
I see your point about the Money.toString method possibly breaking someone's app code.  I do think it's a good idea to have the default toString method show full precision.  

If this does break someone's code, they should be able to fix it using the toString(uom, format) method to provide stricter formatting. I think if we add a migration note to the Release History, that should suffice.

bytecodeguru:
Thanks for the fix 👍

cquiroz:
I have published a snapshot of squants up to this point
We should consider to make another release when the pending PRs get merged

PR #218, opened 2017-04-07
Stabilized toString output for MoneyContext and MoneyNumeric

garyKeorkunian:
* Added custom `toString` methods to `MoneyContext` and `MoneyNumeric` to stabilize their output.  This should prevent thrash in the README file on each run of `tut`.

No tests should be required for these methods, as they are primarily used in the REPL and production application code should generally not depend on these values.

### DIFF
--- a/shared/src/main/scala/squants/market/MoneyContext.scala
+++ b/shared/src/main/scala/squants/market/MoneyContext.scala
@@ -8,6 +8,8 @@
 
 package squants.market
 
+import scala.collection.SortedSet
+
 /**
  * MoneyContext
  *
@@ -23,7 +25,6 @@ package squants.market
  *
  * @author  garyKeorkunian
  * @since   0.1
- *
  * @param defaultCurrency Currency used when none is supplied to the Money factory
  * @param rates Collection of Exchange Rates used for currency conversions
  */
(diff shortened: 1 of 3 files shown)


cquiroz:
This looks good, though I'm not sure about the assumption that it wouldn't impact application code. Often code relies on `toString` for better or worse. Perhaps we should add a note to the Readme/Release notes?

garyKeorkunian:
Yes, I suppose some apps could be using in logs or other ways I've not anticipated.  Once it's merged, I'll add a migration note to the release history.

shared/src/main/scala/squants/market/MoneyContext.scala
```diff
@@ -34,6 +35,26 @@ case class MoneyContext(
     allowIndirectConversions: Boolean = true) {
 
   /**
+    * Custom implementation using SortedSets to ensure consistent output
+    * @return String representation of this instance
+    */
+  override def toString: String = string
```
derekmorr: I wonder if we could just map over `currencies` to get their symbols and make a `SortedSet` from that. And do the same for `rates`. That would avoid needing to create `Ordering` instances for `Currency` and `CurrencyExchangeRate`.

Something like:

```scala
val cSet = currencies.map { _.code }.sorted.mkString(',')
```

garyKeorkunian: Yes, much cleaner.  I needed to through a toList call in the pipeline to provide sortability without the Ordering implicit.  I'll have a new push soon.

garyKeorkunian: The SortedSet was overkill for the internal implementation, for sure.

PR #219, opened 2017-04-24
Compilers and dependency updates

golem131:
Update to sbt 0.13.15
Update to Scala 2.12.2 and 2.11.11
Update sbt to 0.13.15
Update Json4s to 3.5.1
Update scalatest to 3.0.2
Update scalacheck to 1.13.5

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: scala
 
 scala:
    - 2.10.6
-   - 2.11.8
+   - 2.11.11
 
 jdk:
    - oraclejdk7
@@ -10,7 +10,7 @@ jdk:
 
 matrix:
   include:
-    - scala: 2.12.1
+    - scala: 2.12.2
       jdk: oraclejdk8
       env:
         - TRAVIS_NODE_VERSION="6"
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -6,12 +6,12 @@ import com.typesafe.sbt.osgi.SbtOsgi.autoImport._
 
 object Versions {
   val Squants = "1.3.0-SNAPSHOT"
-  val Scala = "2.11.8"
-  val ScalaCross = Seq("2.12.1", "2.11.8", "2.10.6")
+  val Scala = "2.11.11"
+  val ScalaCross = Seq("2.12.2", "2.11.11", "2.10.6")
 
-  val ScalaTest = "3.0.1"
-  val ScalaCheck = "1.13.4"
-  val Json4s = "3.5.0"
+  val ScalaTest = "3.0.3"
+  val ScalaCheck = "1.13.5"
+  val Json4s = "3.5.1"
 }
 
 object Dependencies {
--- a/project/build.properties
+++ b/project/build.properties
@@ -1,1 +1,1 @@
-sbt.version=0.13.13
+sbt.version=0.13.15
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -1,4 +1,4 @@
-addSbtPlugin("org.scala-js" % "sbt-scalajs" % "0.6.14")
+addSbtPlugin("org.scala-js" % "sbt-scalajs" % "0.6.16")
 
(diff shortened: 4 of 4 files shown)


cquiroz:
Thanks, it seems there is a conflict with scalatest and jdk 1.7
I see this on travis
```
java.lang.UnsupportedClassVersionError: org/scalatest/DoNotDiscover : Unsupported major.minor version 52.0
```

golem131:
@cquiroz I'll fix it tomorrow

derekmorr:
I'm concerned about increased CI times with this. It was only recently that we got CI build times under control (in #195 ). This is adding more JDKs back in. When we have multiple outstanding PRs, Travis gets backed up for hours running through builds that, frankly, don't tell us much.

golem131:
@derekmorr i understand you position, but maybe it's time to deprecate support of Scala 2.10 or JDK 7 (all of them EOL now).

derekmorr:
I'd love to kill both 2.10 and 7. But have enough projects moved off it? IIRC, Spark 1.x is on 2.10. In any event, we need to give the community notice that we plan to drop support for 2.10 and 7.

cquiroz:
I'm one of the ones that still needs to use it on 2.10 and jdk 7 😢  
I'd love to upgrade but can't easily

golem131:
@derekmorr @cquiroz anyway, i think, that it's very important to start test build for support JDK 9 and Scala 2.13 as early, as we can.

cquiroz:
I don't disagree, if we can get onboard scala 2.13 early I'm happy

golem131:
@cquiroz @derekmorr and what about JDK 9?

golem131:
JDK 9 + Scala 2.13 add two builds

derekmorr:
It's not just two builds. It's two for the PR and two for the branch. Per-branch and PR. I know it sounds small, but each new build adds up. Many many times I've had Travis get backed up for 3 or 4 hours working through builds for the various combinations. And I don't think I've seen it once give useful output where we had breakage on, say java 7 but not 8, or node 4 but not 6, etc.

golem131:
@derekmorr ok, and in such logic we can just remove jdk 7 builds.

cquiroz:
In that sense giving up on node versions is a smaller risk, squants is more jvm oriented than js oriented, though I personally use it in both

golem131:
@cquiroz we already test only one version of node - 6 (LTS)

cquiroz:
Do we already only test scala 2.12 and up only on jdk 8? so it is only scala 2.10 and 2.11 on jdk 7 right?

golem131:
we test 
* Scala 2.10/JDK7
* Scala 2.11/JDK7
* Scala 2.11/JDK8
* Scala 2.12/JDK8

I want to add Scala 2.12/JDK9
And after publishing json4s build for Scala 2.13.0-M1 add  Scala 2.13.0-M1/JDK8

derekmorr:
We also test on 2.10 + JDK 8.

I'm not sure I see the benefit of testing on 2.13 and JDK 9 this early in their release process. There will likely be many changes in both of them, which could affect us. Once they have release candidates out, I would add them in.

cquiroz:
I guess we can have them on the build.sbt but not put them on travis yet. I personally don't mind either way

derekmorr:
Let me make a suggestion. Let's split the PR into two:

One that updates the existing libraries:

Update to sbt 0.13.15
Update to Scala 2.12.2 and 2.11.11
Update sbt to 0.13.15
Update Json4s to 3.5.1
Update scalatest to 3.0.2
Update scalacheck to 1.13.5

The above versions bumps are non-controversial.

Then have a second PR that adds 2.13 and JDK9 9 where we can have this conversation.

golem131:
@derekmorr - ok, let's try

golem131:
@cquiroz review  this PR, pls

derekmorr:
LGTM.

derekmorr:
Thanks. Merged.

PR #220, opened 2017-04-28
add SiUnit marker trait to SI units in electro package

derekmorr:
This will help #215 .

### DIFF
--- a/shared/src/main/scala/squants/electro/Capacitance.scala
+++ b/shared/src/main/scala/squants/electro/Capacitance.scala
@@ -49,27 +49,27 @@ object Farads extends CapacitanceUnit with PrimaryUnit with SiUnit {
   val symbol = "F"
 }
 
-object Picofarads extends CapacitanceUnit {
+object Picofarads extends CapacitanceUnit with SiUnit {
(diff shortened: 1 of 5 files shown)


cquiroz:
A general question, why `SiUnits` instead of `SIUnits`? SI is an acronym right?

derekmorr:
It is, but `SiUnit` already existed. If we rename it, we break compatibility. 

I don't have strong opinions about it, but, personally, I find `SiUnit` easier to read than `SIUnit` (although that's entirely subjective).

cquiroz:
Is the change in #215 big enough to break compatibility?

cquiroz:
Now I understand how this didn't break the build `SiUnit` is already present. I don't think there is a need to change it if already present with that name

derekmorr:
Right. The `SiUnit` trait already existed. It was inconsistently applied to various units. I'm adding it to units where appropriate, so I can more easily group them together (in #215).

#215 should only add new capabilities. It doesn't change anything.

cquiroz:
Allright, that sounds perfectly fine

PR #222, opened 2017-04-28
add SiUnit trait to SI units in mass package

derekmorr:
This is in service of #215 .

I also fixed a bug where all `DensityUnit`s were annotated with `SiUnit` (`SiUnit` was mixed into the `DensityUnit` base trait).

### DIFF
--- a/shared/src/main/scala/squants/mass/AreaDensity.scala
+++ b/shared/src/main/scala/squants/mass/AreaDensity.scala
@@ -54,7 +54,7 @@ object KilogramsPerHectare extends AreaDensityUnit with UnitConverter {
   val conversionFactor = 1/(100*100d)
 }
 
-object GramsPerSquareCentimeter extends AreaDensityUnit with UnitConverter {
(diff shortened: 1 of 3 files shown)


shared/src/main/scala/squants/mass/AreaDensity.scala
```diff
@@ -54,7 +54,7 @@ object KilogramsPerHectare extends AreaDensityUnit with UnitConverter {
   val conversionFactor = 1/(100*100d)
 }
 
-object GramsPerSquareCentimeter extends AreaDensityUnit with UnitConverter {
+object GramsPerSquareCentimeter extends AreaDensityUnit with UnitConverter with SiUnit {
```
cquiroz: Are you derived units too? I'm not sure if `AreaDensity` has a formal SI definition, Do you have a reference?

derekmorr:
Fair point. I'm not sure exactly how we should handle this. I think Mark other derived units as `SiUnit`, so this shouldn't set a precedent.

derekmorr:
What did we decide to do w/ this ?

cquiroz:
I think this is fine

PR #227, opened 2017-04-30
Rotational dynamics units

Paxelord:
Create units related to rotational dynamics such as Torque, MomentOfInertia, and AngularAcceleration.

### DIFF
--- a/shared/src/main/scala/squants/Quantity.scala
+++ b/shared/src/main/scala/squants/Quantity.scala
@@ -309,5 +309,7 @@ abstract class Quantity[A <: Quantity[A]] extends Serializable with Ordered[A] {
 }
 
 abstract class StrictlyPositiveQuantity[A <: StrictlyPositiveQuantity[A]](value: Double) extends Quantity[A] {self: A ⇒
-  assert(value > 0, "Moment of inertia must be positive.")
+  if(value < 0) {
+    throw new IllegalArgumentException("Cannot create negative StrictlyPositiveQuantity")
+  }
 }
--- a/shared/src/main/scala/squants/Quantity.scala
+++ b/shared/src/main/scala/squants/Quantity.scala
@@ -308,8 +308,3 @@ abstract class Quantity[A <: Quantity[A]] extends Serializable with Ordered[A] {
   def map(f: Double ⇒ Double): A = unit(f(value))
 }
 
-abstract class StrictlyPositiveQuantity[A <: StrictlyPositiveQuantity[A]](value: Double) extends Quantity[A] {self: A ⇒
-  if(value < 0) {
-    throw new IllegalArgumentException("Cannot create negative StrictlyPositiveQuantity")
-  }
-}
--- a/shared/src/main/scala/squants/Quantity.scala
+++ b/shared/src/main/scala/squants/Quantity.scala
@@ -306,5 +306,8 @@ abstract class Quantity[A <: Quantity[A]] extends Serializable with Ordered[A] {
    * @return
    */
   def map(f: Double ⇒ Double): A = unit(f(value))
+}
 
+abstract class StrictlyPositiveQuantity[A <: StrictlyPositiveQuantity[A]](value: Double) extends Quantity[A] {self: A ⇒
+  assert(value > 0, "Moment of inertia must be positive.")
 }
--- a/shared/src/main/scala/squants/mass/MomentOfIntertia.scala
+++ b/shared/src/main/scala/squants/mass/MomentOfIntertia.scala
@@ -2,7 +2,7 @@ package squants.mass
 
 import squants.motion.{AngularAcceleration, NewtonMeters, Torque}
 import squants.space.{Feet, Meters}
-import squants.{Dimension, PrimaryUnit, SiBaseUnit, StrictlyPositiveQuantity, UnitConverter, UnitOfMeasure}
+import squants.{AbstractQuantityNumeric, Dimension, Length, PrimaryUnit, Quantity, SiBaseUnit, UnitConverter, UnitOfMeasure}
 
 /**
   *
@@ -12,7 +12,7 @@ import squants.{Dimension, PrimaryUnit, SiBaseUnit, StrictlyPositiveQuantity, Un
   * @param value Double
   */
 final class MomentOfInertia private (val value: Double, val unit: MomentOfInertiaUnit)
-    extends StrictlyPositiveQuantity[MomentOfInertia](value){
+    extends Quantity[MomentOfInertia]{
 
   def dimension = MomentOfInertia
 
@@ -20,10 +20,19 @@ final class MomentOfInertia private (val value: Double, val unit: MomentOfInerti
   def toPoundsSquareFeet = to(PoundsSquareFeet)
 
   def *(angularAcceleration: AngularAcceleration): Torque = {
-    val kilogramsMetersSquared = toKilogramsMetersSquared
     val radiansPerSecondSquared = angularAcceleration.toRadiansPerSecondSquared
 
-    NewtonMeters(kilogramsMetersSquared * radiansPerSecondSquared)
+    NewtonMeters(toKilogramsMetersSquared * radiansPerSecondSquared)
+  }
+
+  /**
+    * For a point mass with the given MomentOfInertia rotating with a center of
+    * rotation at the given radius, return the mass of the point mass
+    * @param radius distance to axis of rotation
+    * @return mass of point mass with given radius and MomentOfInertia
+    */
+  def atCenter(radius: Length): Mass = {
+    Kilograms(toKilogramsMetersSquared / radius.squared.toSquareMeters)
   }
 }
 
@@ -49,4 +58,16 @@ object KilogramsMetersSquared extends MomentOfInertiaUnit with PrimaryUnit with 
 object PoundsSquareFeet extends MomentOfInertiaUnit {
   val symbol = Pounds.symbol + "‧" + Feet.symbol + "²"
   val conversionFactor = Pounds.conversionFactor * math.pow(Feet.conversionFactor, 2D)
+}
+
+object MomentOfInertiaConversions {
+  lazy val kilogramMetersSquared = KilogramsMetersSquared(1)
+  lazy val poundSquareFeet = PoundsSquareFeet(1)
+
+  implicit class MomentOfInertiaConversions[A](val n: A) extends AnyVal {
+    def kilogramMetersSquared(implicit num: Numeric[A]) = KilogramsMetersSquared(n)
+    def poundSquareFeet(implicit num: Numeric[A]) = PoundsSquareFeet(n)
+  }
+
+  implicit object MomentOfInertiaNumeric extends AbstractQuantityNumeric[MomentOfInertia](MomentOfInertia.primaryUnit)
 }
--- a/shared/src/main/scala/squants/mass/MomentOfIntertia.scala
+++ b/shared/src/main/scala/squants/mass/MomentOfIntertia.scala
@@ -4,13 +4,20 @@ import squants.motion.{AngularAcceleration, NewtonMeters, Torque}
 import squants.space.{Feet, Meters}
 import squants.{Dimension, PrimaryUnit, SiBaseUnit, StrictlyPositiveQuantity, UnitConverter, UnitOfMeasure}
 
+/**
+  *
+  * @author paxelord
+  * @since 1.2
+  *
+  * @param value Double
+  */
 final class MomentOfInertia private (val value: Double, val unit: MomentOfInertiaUnit)
-  extends StrictlyPositiveQuantity[MomentOfInertia](value){
+    extends StrictlyPositiveQuantity[MomentOfInertia](value){
 
   def dimension = MomentOfInertia
 
-  def toKilogramsMetersSquared: Double = to(KilogramsMetersSquared)
-  def toPoundsSquareFeet: Double = to(PoundsSquareFeet)
+  def toKilogramsMetersSquared = to(KilogramsMetersSquared)
+  def toPoundsSquareFeet = to(PoundsSquareFeet)
 
   def *(angularAcceleration: AngularAcceleration): Torque = {
     val kilogramsMetersSquared = toKilogramsMetersSquared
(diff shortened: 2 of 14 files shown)


shared/src/main/scala/squants/Quantity.scala
```diff
@@ -306,5 +306,10 @@ abstract class Quantity[A <: Quantity[A]] extends Serializable with Ordered[A] {
    * @return
    */
   def map(f: Double ⇒ Double): A = unit(f(value))
+}
 
+abstract class StrictlyPositiveQuantity[A <: StrictlyPositiveQuantity[A]](value: Double) extends Quantity[A] {self: A ⇒
```
cquiroz: This is interesting though I'd strongly prefer have a constructor returning an `Option[StrictlyPositiveQuantity[A]]` than throwing an exception here

shared/src/main/scala/squants/mass/MomentOfIntertia.scala
```diff
@@ -0,0 +1,73 @@
+package squants.mass
+
+import squants.motion.{AngularAcceleration, NewtonMeters, Torque}
+import squants.space.{Feet, Meters}
+import squants.{AbstractQuantityNumeric, Dimension, Length, PrimaryUnit, SiBaseUnit, StrictlyPositiveQuantity, UnitConverter, UnitOfMeasure}
+
+/**
+  *
+  * @author paxelord
+  * @since 0.6
```
cquiroz: We are at version 1.2

shared/src/main/scala/squants/mass/MomentOfIntertia.scala
```diff
@@ -0,0 +1,73 @@
...
+object PoundsSquareFeet extends MomentOfInertiaUnit {
+  val symbol = Pounds.symbol + "‧" + Feet.symbol + "²"
+  val conversionFactor = Pounds.conversionFactor * math.pow(Feet.conversionFactor, 2D)
+}
+
+object MomentOfInertiaConversions {
+  lazy val kilogramMetersSquared = KilogramsMetersSquared(1)
+  lazy val poundSquareFeet = PoundsSquareFeet(1)
+
+  implicit class MassConversions[A](n: A)(implicit num: Numeric[A]) {
```
cquiroz: Can this be made a value class? maybe not with the implicit there

shared/src/main/scala/squants/mass/MomentOfIntertia.scala
```diff
@@ -0,0 +1,73 @@
...
+}
+
+trait MomentOfInertiaUnit extends UnitOfMeasure[MomentOfInertia] with UnitConverter {
+  def apply[A](n: A)(implicit num: Numeric[A]) = {
+    MomentOfInertia(num.toDouble(n), this)
+  }
+}
+
+object KilogramsMetersSquared extends MomentOfInertiaUnit with PrimaryUnit with SiBaseUnit {
+  val symbol = Kilograms.symbol + "‧" + Meters.symbol + "²"
```
cquiroz: Are we using exponentials in unicode like this in other places?

shared/src/main/scala/squants/motion/AngularAcceleration.scala
```diff
@@ -0,0 +1,114 @@
...
+                                      UnitConverter {
+  def apply[A](n: A)(implicit num: Numeric[A]) = {
+    AngularAcceleration(num.toDouble(n), this)
+  }
+
+  val conversionFactor: Double
+}
+
+object RadiansPerSecondSquared extends AngularAccelerationUnit with PrimaryUnit with SiUnit{
+  override val symbol: String = "rad/s²"
```
cquiroz: Wouldn't you use `Radians.symbol` here to be consistent?

shared/src/main/scala/squants/motion/AngularAcceleration.scala
```diff
@@ -0,0 +1,114 @@
...
+}
+
+object DegreesPerSecondSquared extends AngularAccelerationUnit {
+  val symbol = "°/s²"
```
cquiroz: Same as above

shared/src/main/scala/squants/mass/MomentOfIntertia.scala
```diff
@@ -0,0 +1,73 @@
...
+    * @param radius distance to axis of rotation
+    * @return mass of point mass with given radius and MomentOfInertia
+    */
+  def atCenter(radius: Length): Mass = {
+    Kilograms(toKilogramsMetersSquared / radius.squared.toSquareMeters)
+  }
+}
+
+object MomentOfInertia extends Dimension[MomentOfInertia] {
+  private[mass] def apply[A](n: A, unit: MomentOfInertiaUnit)(implicit num: Numeric[A]) = new MomentOfInertia(num.toDouble(n), unit)
```
cquiroz: Could the constructors here return an `Option` to account for negative values?

shared/src/main/scala/squants/motion/AngularAcceleration.scala
```diff
@@ -0,0 +1,114 @@
...
+  def *(that: MomentOfInertia): Torque = {
+    NewtonMeters(toRadiansPerSecondSquared * that.toKilogramsMetersSquared)
+  }
+
+  override protected[squants] def timeIntegrated: AngularVelocity = RadiansPerSecond(toRadiansPerSecondSquared)
+
+  override protected[squants] def time: Time = Seconds(1)
+}
+
+
```
cquiroz: 2 lines?

shared/src/main/scala/squants/motion/AngularVelocity.scala
```diff
@@ -20,16 +20,28 @@ import squants.time.TimeDerivative
...
   def toGradsPerSecond = to(GradsPerSecond)
   def toTurnsPerSecond = to(TurnsPerSecond)
 
+  /**
+    * linear velocity of an object rotating with this angular velocity
+    * and the given radius from the center of rotation
+    * @param radius the distance from the center of rotation
+    * @return linear velocity with given angular velocity and radius
+    */
+  def onRadius(radius: Length): Velocity = {
```
cquiroz: You don't need the `{` parentheses here

shared/src/main/scala/squants/motion/Torque.scala
```diff
@@ -0,0 +1,61 @@
...
+}
+
+trait TorqueUnit extends UnitOfMeasure[Torque] with UnitConverter {
+  def apply[A](n: A)(implicit num: Numeric[A]) = {
+    Torque(num.toDouble(n), this)
+  }
+}
+
+object NewtonMeters extends TorqueUnit with PrimaryUnit with  SiBaseUnit {
+  val symbol = Newtons.symbol + "‧" + Meters.symbol
```
cquiroz: Is this  a `times` symbol? Are we using the same on other units?

shared/src/main/scala/squants/mass/MomentOfIntertia.scala
Paxelord: In Acceleration, unicode exponentials are used.

shared/src/main/scala/squants/mass/MomentOfIntertia.scala
Paxelord: I'm not 100% sure that's a good idea. People might be confused why they are getting a `None.get` Exception, whereas as throwing an IlleagalArgumentException would be more descriptive.

shared/src/main/scala/squants/motion/AngularAcceleration.scala
Paxelord: My bad, I'll change that!

shared/src/main/scala/squants/mass/MomentOfIntertia.scala
derekmorr: Throwing exceptions isn't very functional. If `Option` isn't suitable, we can look at `Either` which has a place to place an error data structure.

shared/src/main/scala/squants/motion/Torque.scala
Paxelord: This is meant to be the times symbol. I looked though squants, and I think this is the first time we have a times symbol in a unit. It might be clearer to replace the current bullet `‧` with a splat `*`

In a future PR, it might be useful to have have a `Product[Unit]` to allow for easy extendability of units. For example, `MomentOfInertia` would extend `Product[Area]` (because `Area` is `Length` squared) and `Product[Mass]` But that's for a future PR.

shared/src/main/scala/squants/mass/MomentOfIntertia.scala
Paxelord: I don't understand, can you please clarify?

If it helps, I accidentally misnamed what should have been `MomentOfInertiaConversions` to `MassConversions` by mistake.

shared/src/main/scala/squants/mass/MomentOfIntertia.scala
Paxelord: I will make sure to change this

shared/src/main/scala/squants/motion/AngularAcceleration.scala
Paxelord: I will make sure to change this to just 1 line.

shared/src/main/scala/squants/mass/MomentOfIntertia.scala
Paxelord: I will try experimenting with `Option` and `Either`, and try to figure out what is most natural

Paxelord: On a side note, I am starting to second guess whether this class should even exist. Maybe Squants should only provide compile time safety of units, and not tell people how to represent quantities, even if they are doing unusual things like having negative `Mass` or `MomentOfInertia`

shared/src/main/scala/squants/mass/MomentOfIntertia.scala
shadaj: Just chiming in, @Paxelord and I are on the same robotics team, where we're using Squants. I think this can still be made into a value class, if the implicit is moved to the methods. So `MassConversions` (or `MomentofInertiaConversions` I guess) would only take a `val n`, and the `kilogramMetersSquared` and `poundSquareFeet` would take the implicit `Numeric`.

shared/src/main/scala/squants/mass/MomentOfIntertia.scala
Paxelord: One compromise would to have the strictly positive feature optional, enabled by default. Someone who needs to use negative mass would have to explicitly disable the strictly positive feature. I'm not completely sure how that would work though.

shared/src/main/scala/squants/Quantity.scala
Paxelord: @cquiroz this feature leads to many different complications that are outside of the scope of this PR, so I have decided to remove it. I have addressed all your  comments, and I think this PR is now ready to be merged! Thank you!

shared/src/main/scala/squants/motion/AngularAcceleration.scala
```diff
@@ -58,15 +67,15 @@ trait AngularAccelerationUnit extends UnitOfMeasure[AngularAcceleration] with
...
 }
 
 object DegreesPerSecondSquared extends AngularAccelerationUnit {
-  val symbol = "°/s²"
+  val symbol = Degrees.symbol + "/s²"
   val conversionFactor = Degrees.conversionFactor
 }
 
-object GradiansPerSecondSquared extends AngularAccelerationUnit {
+object GradsPerSecondSquared extends AngularAccelerationUnit {
```
cquiroz: Why the change to `Grads`? `Gradians` seems clearer at the costs of just a few letters. In Spanish `Grads` can be easily confused by `Degrees`

shared/src/main/scala/squants/Quantity.scala
```diff
@@ -308,8 +308,3 @@ abstract class Quantity[A <: Quantity[A]] extends Serializable with Ordered[A] {
   def map(f: Double ⇒ Double): A = unit(f(value))
 }
 
-abstract class StrictlyPositiveQuantity[A <: StrictlyPositiveQuantity[A]](value: Double) extends Quantity[A] {self: A ⇒
```
cquiroz: Sorry about this. I think this can be made to work just making the constructors return `Option` or `Either` or even `Try`

cquiroz:
I'm ok with this PR as it is now but will wait for others to chime in, @derekmorr?

shared/src/main/scala/squants/motion/AngularAcceleration.scala
Paxelord: I made it `Grads` instead of `Gradians` because that was the convention used in `AngularVelocity`. However, I agree with you about `Gradians`  being clearer, and I will change it. Do you also want to change that in `AngularVelocity`?

cquiroz: Yes, I think it should be consistent with the units on `Angle`

https://github.com/typelevel/squants/blob/master/shared/src/main/scala/squants/space/Angle.scala#L66

shared/src/main/scala/squants/Quantity.scala
Paxelord: The problem is that we run into complications such as these: Take the scenario `KilogramsMetersSquared(1) - KilogramsMetersSquared(2) + KilogramsMetersSquared(2)`
After the first subtraction, we get a negative moment, resulting in `None`. The addition afterwards would add a `None` with a `KilogramsMetersSquared(2)`, getting `None`
However, mathematically, this should be evaluated to the equivalent of `Some(KilogramsMetersSquared(1))`.  Implementing this kind of feature is certainly possible, but scenarios such as these make it complicated and outside the scope of this pr.

cquiroz:
Can we proceed with this PR? It seems fine to me

derekmorr:
My only objection was renaming `GradsPerSecond` to `GradiansPerSecond` since it changes existing API.

cquiroz:
I see, I proposed that to make it consistent with `Angle` but right it would break the API. Could we deprecate the existing and add `Gradians`?

derekmorr:
Sure that would work. Do you want to add that to this PR or make a new one for it ?

cquiroz:
@Paxelord Could you implement the change suggested by @derekmorr and then we can merge

Paxelord:
@cquiroz I have added the `GradsPerSecond` back, in addition to deprecation warnings.

PR #234, opened 2017-05-12
Unit groups basic definition

derekmorr:
This is the first (of several) PRs for #212. It has the basic trait for groups `UnitOfMeasure` that everything else will build on.


cquiroz:
Thanks for fixing typos along the way 👍

PR #235, opened 2017-05-12
Add SI, Troy, US Customary unit groups

derekmorr:
This builds on #234 to start adding unit groups.

There are two SI unit groups, "strict" and "expanded." Strict is just those in the SI itself. Expanded adds hours, minutes, days, liters, and other non-SI units that can be used in the SI, as defined in the [BIPM appendix](http://www.bipm.org/en/publications/si-brochure/table6.html).

I defined the SI using implicits to avoid having a massive file with lots of boilerplate calling `... collect SiUnit si ...` for each `UnitOfMeasure`. In order to get `UnitGroup`s implicitly generated, I had to get `Dimension` into implicit scope, which I did via the `ImplicitDimensions` object. That's just boilerplate, but at least it's only one per Dimension.

### DIFF
--- a/shared/src/main/scala/squants/unitgroups/ImplicitDimensions.scala
+++ b/shared/src/main/scala/squants/unitgroups/ImplicitDimensions.scala
@@ -1,0 +1,104 @@
+package squants.unitgroups
+
+import squants.Dimension
+import squants.electro._
+import squants.energy._
+import squants.information.{DataRate, Information}
+import squants.mass.{AreaDensity, ChemicalAmount, Density, Mass}
+import squants.motion._
+import squants.photo._
+import squants.radio._
+import squants.space._
+import squants.thermal.{Temperature, ThermalCapacity}
+import squants.time.{Frequency, Time}
+
+/**
+  * Puts [[Dimension]] objects into implicit scope
+  * so that implicit [[UnitGroup]] machinery works.
+  */
+object ImplicitDimensions {
+
+  object electro {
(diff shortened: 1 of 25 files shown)


derekmorr:
i have another commit pending for this that adds the new objects from #227 .

shared/src/main/scala/squants/unitgroups/ImplicitDimensions.scala
```diff
@@ -0,0 +1,104 @@
...
+import squants.radio._
+import squants.space._
+import squants.thermal.{Temperature, ThermalCapacity}
+import squants.time.{Frequency, Time}
+
+/**
+  * Puts [[Dimension]] objects into implicit scope
+  * so that implicit [[UnitGroup]] machinery works.
+  */
+object ImplicitDimensions {
```
cquiroz: Just a general were question. If we were going to add a new `Quantity` would there be compilation errors if the implicit `Dimension` isn't defined here? I hope that's the case 😄

derekmorr: No, there wouldn't be. You'll get a compiler error if you try to use it, but this file will build just fine. Which is a problem. I'm not sure the best way to address it.

derekmorr: We'd have the same issue if we wrote lots of boilerplate collecting `SiUnit` for each `Dimension` too, though.

PR #238, opened 2017-05-14
Add Scala Native as a supported platform

shadaj:
This adds cross-compilation of Squants to Scala Native (for Scala 2.11 only). There is a small hack in `toString` for `Quantity` to work around string formatting being unimplemented in Scala Native. This also updates publishing instructions to handle Scala Native only supporting Scala 2.11.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1224,9 +1224,11 @@ To make a release do the following:
   sbt tut
 ```
 
-* Publish a cross-version signed package
+* Publish a cross-version signed package (no cross-version available for Scala Native)
 ```
-  sbt +publishSigned
+  sbt +squantsJVM/publishSigned
+  sbt +squantsJS/publishSigned
+  sbt squantsNative/publishSigned
 ```
 
 * Then make a release (Note: after this step the release cannot be replaced)
--- a/shared/src/main/scala/squants/package.scala
+++ b/shared/src/main/scala/squants/package.scala
@@ -118,12 +118,4 @@ package object squants {
     def /(that: Time) = Each(bd) / that
     def per(that: Time): Frequency = /(that)
   }
-
-  /**
-   * Helper function to achieve uniform Double formatting over JVM and JS platforms.
-   * Simple Double.toString will format 1.0 as "1.0" on JVM and as "1" on JS
-   * @param d Double number to be formatted
-   * @return
-   */
-  private[squants] def crossFormat(d: Double): String = if (d.toLong == d) { "%.1f".format(d) } else { d.toString }
 }
--- a/shared/src/main/scala/squants/thermal/Temperature.scala
+++ b/shared/src/main/scala/squants/thermal/Temperature.scala
@@ -9,6 +9,7 @@
 package squants.thermal
 
 import squants._
+import squants.Platform.crossFormat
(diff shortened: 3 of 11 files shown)


cquiroz:
Wow this is great. I'm not too familiar with `scala-native` tooling. Can you run tests there?

shadaj:
Not at the moment; Scala Native doesn't have support for running tests from SBT, and ScalaTest and ScalaCheck are not available there yet. So the support is very experimental.

shared/src/main/scala/squants/package.scala
```diff
@@ -125,5 +125,16 @@ package object squants {
    * @param d Double number to be formatted
    * @return
    */
-  private[squants] def crossFormat(d: Double): String = if (d.toLong == d) { "%.1f".format(d) } else { d.toString }
+  private[squants] def crossFormat(d: Double): String = {
```
cquiroz: On `scala.js` we can put different implementations of a class for each platform under the `js` and `jvm` folders. Can we do the same and put the native specific code is a separate location?

shadaj: I added a `Platform` object that contains platform-specific implementations of `crossFormat`

shared/src/main/scala/squants/thermal/Temperature.scala
```diff
@@ -9,6 +9,7 @@
 package squants.thermal
 
 import squants._
+import squants.Platform.crossFormat
```
cquiroz: Interesting, is this really the only place this is being used? Regardless I like this solution

cquiroz:
This is looking good and I'm all for including this new capability it but I'll wait a while for other members to chime in

derekmorr:
This looks fine to me, and I'm glad can isolate scala native code in its own directory.

README.md
```diff
@@ -1224,9 +1224,11 @@ To make a release do the following:
   sbt tut
 ```
 
-* Publish a cross-version signed package
+* Publish a cross-version signed package (no cross-version available for Scala Native)
 ```
-  sbt +publishSigned
+  sbt +squantsJVM/publishSigned
```
derekmorr: Is it necessary to break these out into three commands now?

cquiroz: That's a good point, breaking it will probably slow down travis a bit while warming up sbt

cquiroz:
@Paxelord Let us know if you can change the travis task split and if you want you can add yourself to the contributors list on the README

If you don't want to change those we can merge and change it later

Thanks a lot for this, I'll start adding scala-native support to my own projects

derekmorr:
He should already be in the contributor list. I added him in #237 , but I forgot to run tut to rebuild the main `README.md`. I believe that's fixed in #239 .

shadaj:
I split up the Travis task because Scala Native only supports Scala 2.11, so we cannot run the cross-version commands for native compilation. As soon as support is added in Scala Native for 2.10 and 2.12 the change can be reverted run cross-version commands for all platforms.

Also, just a little confusion, @Paxelord is another member on the same robotics team as me (@Team846). But I pinged him to let him know about the contributors list since he contributed additional angular units in a previous PR.

PR #240, opened 2017-05-17
Formatters

derekmorr:
This adds formatters for inBestUnit. Docs will come in a later commit.

### DIFF
--- a/shared/src/main/scala/squants/formatter/DefaultFormatter.scala
+++ b/shared/src/main/scala/squants/formatter/DefaultFormatter.scala
@@ -1,0 +1,21 @@
+package squants.formatter
+
+import squants.Quantity
+import squants.unitgroups.UnitGroup
+
+trait DefaultFormatter[A <: Quantity[A]] extends Formatter[A] {
+  def unitGroup: UnitGroup[A]
+
+  override def bestUnit(quantity: Quantity[A]): A = {
+    if (unitGroup.units.isEmpty)
+      quantity.in(quantity.unit)
+    else {
+      val unit =
+        unitGroup.sortedUnits
+          .takeWhile { u => quantity.to(u).abs >= 1.0 }
+          .lastOption
+          .getOrElse(unitGroup.sortedUnits.head)
+      quantity.in(unit) 
+    }
+  }
+}
--- a/shared/src/test/scala/squants/formatter/DefaultFormatterSpec.scala
+++ b/shared/src/test/scala/squants/formatter/DefaultFormatterSpec.scala
@@ -1,0 +1,33 @@
+package squants.formatter
+
+import org.scalatest.{FlatSpec, Matchers}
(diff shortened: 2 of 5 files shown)


shared/src/main/scala/squants/formatter/DefaultFormatter.scala
```diff
@@ -0,0 +1,22 @@
...
+
+import squants.Quantity
+import squants.unitgroups.UnitGroup
+
+
+trait DefaultFormatter[A <: Quantity[A]] extends Formatter[A] {
+  def unitGroup: UnitGroup[A]
+
+  override def bestUnit(quantity: Quantity[A]): A = {
+    if (unitGroup.units.isEmpty)
```
cquiroz: Can this be pattern matched rather than using an if?

shared/src/test/scala/squants/formatter/DefaultFormatterSpec.scala
```diff
@@ -0,0 +1,34 @@
...
+
+  val usLengthFormatter = new DefaultFormatter[Length] { val unitGroup = UsCustomaryLengths }
+
+  behavior of "DefaultFormatter"
+
+  it should "pick the best unit - when the unit stays the same" in {
+    usLengthFormatter.bestUnit(5.inches).unit should be (Inches)
+  }
+
+
```
cquiroz: double line?

shared/src/main/scala/squants/formatter/DefaultFormatter.scala
```diff
@@ -0,0 +1,22 @@
+package squants.formatter
+
+import squants.Quantity
+import squants.unitgroups.UnitGroup
+
+
```
cquiroz: double lines?

derekmorr: fixed.

shared/src/main/scala/squants/formatter/DefaultFormatter.scala
derekmorr: I'm not sure. `Set.empty` isn't a case object like `Nil`, so pattern matching on that won't work.

shared/src/test/scala/squants/formatter/DefaultFormatterSpec.scala
derekmorr: fixed

shared/src/main/scala/squants/formatter/DefaultFormatter.scala
cquiroz: Right, don't worry

derekmorr:
Are we good to merge this? I can do docs and REPL updates in a second PR, like I did for UnitGroups, if that's ok.

cquiroz:
Seems fine to me

derekmorr:
Ok, merged.

PR #244, opened 2017-05-24
add formatter docs; minor adjustments for formatter syntax

derekmorr:
This adds docs for formatters. I made some changes since the last PR to make the syntax more consistent.

I'd appreciate feedback on the docs. It can get complicated with unitgroups, formatters, implicits, etc. I'm not sure I structured it in the best way.


shared/src/main/tut/README.md
```diff
@@ -709,6 +709,123 @@ The `UnitGroup` values provided with Squants are only samples and aren't intende
 We encourage users to make their own `UnitGroup` defintitions and submit them as PRs if they're generally
 applicable.
 
+## Formatters
+
+Squants provides an experimental API for formatting Quantities in the "best unit." For example,
```
cquiroz: I wonder a bit about the `experimental` part. If this is something we may change then the `experimental` tag fits very well but then perhaps it should be on an `experimental` package.

Do you foresee possible changes in the near future that could break the API?

shared/src/main/tut/README.md
```diff
@@ -709,6 +709,123 @@ The `UnitGroup` values provided with Squants are only samples and aren't intende
...
+To use `DefaultFormatter` import it, and a unit group:
+
+```tut:silent
+import squants.formatter.DefaultFormatter
+import squants.unitgroups.misc.AstronomicalLengthUnitGroup
+```
+
+Then create the formatter by passing in a unit group:
+```tut:book
+val astroFormatter = new DefaultFormatter[Length] { val unitGroup = AstronomicalLengthUnitGroup }
```
cquiroz: Why this isn't built just as a regular class? `new DefaultFormatter[Length](AstronomicalLengthUnitGroup)`

shared/src/main/tut/README.md
derekmorr: I'm not sure. We've not had any feedback on the API from people who requested it. Once it's released, it's possible we'll want to change it based on feedback.

I'm fine moving unitgroups and formatters into an experimental package.

shared/src/main/tut/README.md
derekmorr: Probably b/c everything else was a trait. I agree the class syntax is better.

shared/src/main/tut/README.md
cquiroz: I guess either way the API will change but the experimental would at least be an indication.
I don't have a strong opinion either way

derekmorr: I'll refactor it into an experimental package and update the README.

derekmorr:
I pushed up a commit that addresses the latest comments.

cquiroz:
that seems good, we can move it to non `experimental` on the next release

derekmorr:
Are we ok to merge this then?

cquiroz:
I think so

PR #246, opened 2017-05-25
fix broken scaladoc links

derekmorr:
We had some scaladoc warnings about broken links. This fixes them.


derekmorr:
Hmm. Looks like GitHub won't let me merge unless I make Travis run. I was hoping to avoid that, since it's just a doc fix. I'll push another commit that removes the "[ci skip]" tag.

Issue #243, opened 2017-05-21
Inclusive variant of QuantityRange


derekmorr:
We should have an inclusive variant of `QuantityRange` (i.e., one that includes its upper bound). Currently this is not the case:

```scala
scala> val diskSizes = QuantityRange(10.gib, 500.gib)
diskSizes: squants.QuantityRange[squants.information.Information] = QuantityRange(10.0 GiB,500.0 GiB)

scala> diskSizes.contains(500.gib)
res8: Boolean = false
```

We should also add an appropriate function to `Quantity` to construct inclusive ranges. Unfortunately, this presents a backward-compatibilty problem. In the scala standard library, `to` creates an _inclusive_ range and `until` creates an _exclusive_ range. Currently, `Quantity` has a `to` function that creates an exclusive range (the opposite of the standard library). Ideally, we should change `Quantity`'s behavior to match the standard library, but that risks breaking existing code.

derekmorr:
This isn't needed. I didn't realize that `.contains()` and `.includes()` have different behavior. I updated the docs to reflect this in #247.

PR #247, opened 2017-05-25
update QuantityRange's error message and docs to clarify behavior

### LINKED ISSUE #242: QuantityRange throws spurious exceptions
`QuantityRange` requires that `upper` is strictly greater than `lower`, but the exception it throws claims upper must be >= lower:

```scala
case class QuantityRange[A <: Quantity[A]](lower: A, upper: A) {
  if (lower >= upper) throw new IllegalArgumentException("QuantityRange upper bound must be greater than or equal to the lower bound")
```

We should either fix the conditional check or the error message.

For reference, the Scala standard library, doesn't work when lower == upper:

```scala
scala> val r = Range(1,1)
r: scala.collection.immutable.Range = Range()

scala> r.contains(1)
res4: Boolean = false
```

So perhaps we should leave the conditional in place and change the error message.

I'd like to get this fixed as part of 1.3.0.

derekmorr:
This closes #242 and #243 .

It turns out `QuantityRange` already had the inclusive behavior I wanted, so I updated the README and ScalaDoc to clarify that. I also changed the wording in the constructor's exception message to correctly describe the check.

In a later PR (post-1.3.0) we can add safe building functions that return an Option or Either instead of throwing exceptions.

### DIFF
--- a/shared/src/main/scala/squants/QuantityRange.scala
+++ b/shared/src/main/scala/squants/QuantityRange.scala
@@ -21,7 +21,9 @@ import scala.annotation.tailrec
  * @tparam A the Quantity Type
  */
 case class QuantityRange[A <: Quantity[A]](lower: A, upper: A) {
-  if (lower >= upper) throw new IllegalArgumentException("QuantityRange upper bound must be greater than or equal to the lower bound")
+  if (lower >= upper) {
+    throw new IllegalArgumentException("QuantityRange upper bound must be strictly greater than to the lower bound")
+  }
 
   /**
    * Create a Seq of `multiple` ranges equal in size to the original with sequential range values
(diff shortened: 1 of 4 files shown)


shared/src/main/scala/squants/QuantityRange.scala
```diff
@@ -21,7 +21,9 @@ import scala.annotation.tailrec
  * @tparam A the Quantity Type
  */
 case class QuantityRange[A <: Quantity[A]](lower: A, upper: A) {
-  if (lower >= upper) throw new IllegalArgumentException("QuantityRange upper bound must be greater than or equal to the lower bound")
+  if (lower >= upper) {
```
cquiroz: I wish we were not to throw exceptions on constructors. Instead I'd rather have an `apply` constructor returning an `Option` or `Either` but I guess that would break the API

derekmorr: It would. We can add alternate builder functions. We can do that for 1.3 if you want, or hold off until afterwards.

PR #255, opened 2017-06-06
add ether and litecoin `Currency`s

emanresusername:
i know adding every new crypto currency that pops up would be a full time job :sweat_smile: 
but i figured if coinbase has them they're worth including?


derekmorr:
Thanks for the patch. This looks like it applies cleanly.

Out of curiosity, is there an advantage to including these currencies in the core library? It seems they could be provided as an add-on package. For example, this works in the squants console:

```scala
object CryptoCurrencies {
  object ETH extends Currency("ETH", "Ether", "\u039e", 15)
  object LTC extends Currency("LTC", "Litecoin", "\u0141", 15)
}

implicit class CryptoCurrencyConversions[A](n: A)(implicit num: Numeric[A]) {
  def ETH = Money(n, CryptoCurrencies.ETH)
  def LTC = Money(n, CryptoCurrencies.LTC)
  def ether = ETH
  def liteCoin = LTC
}

import CryptoCurrencies._

val exchangeRates = List(LTC / USD(29.71481364), ETH / USD(254.91))
implicit val ctx = MoneyContext(USD, Set(USD, ETH, LTC), exchangeRates)

50.USD in LTC
// res1: squants.market.Money = 1.682662412282253 LTC
```

emanresusername:
Advantage is proximally just satisfying my weekend whim :sweat_smile: 
(Although i'd imagine working with those denominations would of interest to a larger audience than just me)
I was doing [a little fx project](https://github.com/emanresusername/squants-fx)  to play with this and a few other technologies i'd had my eye on and was pulling in the [open exchange rates](https://docs.openexchangerates.org/docs/latest-json) feed and it had those 2 in it, and i had an interest in them, and here we are :)
Am i reinventing the wheel there? is there already like an existing community `squants-contrib` or `squants-ext` project somewhere out there on the interwebz that i missed?

derekmorr:
No, I don't think there's an existing contrib library. This could be a way to start one. I could go either way on this. @cquiroz what do you think ?

cquiroz:
I think that would be fine. We could probably write a script that could generate all the currencies at once. It's always strange that we support only a few currencies

derekmorr:
I'm not sure we need to support all of them, since it's easy for users to add new ones. I don't really like that users have to keep submitting PRs to get units added -- it should be easier for them to incorporate new units without modifying the library. The market package is one of the few areas where that's true.

Having said that, I'm indifferent on this PR, so I guess we can merge it.

derekmorr:
I'm sorry if "indifferent" sounded too negative - that wasn't my intent.

emanresusername:
s'all good

derekmorr:
Thanks for the PR. I've merged it, and I'll add you to the contributors list.

PR #256, opened 2017-06-06
Update to scalajs 0.6.17

cquiroz:
Update to the latest version of scala.js. This should speed up the creation of tests


derekmorr:
My only concern is that it breaks compatibility:

> However, it is not forward compatible: libraries compiled with 0.6.17 cannot be used by projects using 0.6.{0-16}.
https://www.scala-js.org/news/2017/06/04/announcing-scalajs-0.6.17/

I did some informal benchmarking with 0.6.17 over the weekend, and didn't see much difference in tests.

cquiroz:
it should be faster to build the test suite. as our test suite isn't that large the difference maybe small. In at least one project of mine the difference is very large, but then the test suite is very big

I don't know about the compatibility, that's a good point I guess

derekmorr:
I've seen other project start to upgrade to this version, so I'll go ahead and merge it.

PR #258, opened 2017-06-07
update .gitignore

underscorenico:
I added some general git ignore rules. You can validate and / or check these rules at https://www.gitignore.io/
I was also tempted to add *.sc (scala worksheets), but I saw that .sc files so I left them.


cquiroz:
These seem sensible enough

derekmorr:
These look good to me.

PR #259, opened 2017-06-08
Added Inductance units

underscorenico:
This PR adds four new units to the inductance object (mH, uH, nH, pH).


shared/src/main/scala/squants/electro/Inductance.scala
```diff
@@ -44,11 +48,39 @@ object Henry extends InductanceUnit with PrimaryUnit with SiUnit {
   val symbol = "H"
 }
 
+object Millihenry extends InductanceUnit with SiUnit {
+  val symbol = "mH"
+  val conversionFactor = MetricSystem.Milli
+}
+
+object Microhenry extends InductanceUnit with SiUnit {
+  val symbol = "μH"
```
cquiroz: Are these the official ways to name these units?

underscorenico: Yes they are, at least until a few years ago when I was studying electromagnetism xD

derekmorr:
Thanks for adding the units this.

Would it be possible to split this into two PRs -- one for refactoring (which I'm not convinced of) and one for the units?

underscorenico:
Done!

cquiroz:
@derekmorr Sorry I didn't check your comment about the refactoring and I merged it. Please let's discuss it further and we can revert if neded

derekmorr:
My issue with the `this` changes was that it didn't fix a bug, didn't make illegal state easier to prevent, didn't make future changes easier, and it touched a lot of files. Personally, I'm not a fan of the `this.foo` OO-style convention, but I can put personal preferences aside. My real objection were the other issues.

cquiroz:
Sorry about that, it seemed clearer to me with `this` but I understand your position. If you prefer to revert it, I'm more than fine with that

underscorenico:
This PR only contains one commit: add inductance units and conversions.
As @derekmorr asked, I splitted the original PR into two: this one (only containing the new units) and a new one #260 (containing only the refactoring).

Issue #252, opened 2017-05-31
implement unimplemented functions & make new types


derekmorr:
We have several undefined functions and partial functions in squants. We should fix these for the next release:

```
~/squants/shared/src/main/scala/squants % grep -r "???" .
./electro/Capacitance.scala:  def /(that: Length) = ??? // returns Permittivity
./electro/ElectricCharge.scala:  def /(that: Length) = ??? // returns LinearElectricChargeDensity
./electro/ElectricCharge.scala:  def /(that: Area) = ??? // returns AreaElectricChargeDensity
./electro/ElectricCharge.scala:  def /(that: Volume) = ??? // returns ElectricChargeDensity
./electro/ElectricCharge.scala:  def /(that: Mass) = ??? // returns ElectricChargeMassRatio
./electro/ElectricCurrent.scala:  def /(that: Length) = ??? // returns MagneticFieldStrength
./electro/ElectricCurrent.scala:  def /(that: Area) = ??? // returns ElectricCurrentDensity
./electro/ElectricPotential.scala:  def /(that: Length) = ??? // returns ElectricFieldStrength
./electro/Inductance.scala:  def /(that: Length) = ??? // returns Permeability
./energy/Energy.scala:  def /(that: ChemicalAmount) = ??? // return MolarEnergy
./energy/Energy.scala:  def /(that: Area) = ??? // Insolation, Energy Area Density
./energy/Power.scala:  def /(that: Volume) = ??? // Power Density
./energy/SpecificEnergy.scala:  def /(that: Time) = ??? // returns AbsorbedEnergyRate
./market/Money.scala:  def primaryUnit = ??? // Should not be used with Money - drawn from MoneyContext instead
./market/Money.scala:  def siUnit = ??? // Should not be used with Money - drawn from MoneyContext instead
./market/Money.scala:  def units = ??? // Should not be used with Money - drawn from MoneyContext instead
./market/Money.scala:  protected def converterFrom: Double ⇒ Double = ???
./market/Money.scala:  protected def converterTo: Double ⇒ Double = ???
./mass/ChemicalAmount.scala:  def /(that: Volume) = ??? // returns SubstanceConcentration
./motion/Force.scala:  def /(that: Length) = ??? // return SurfaceTension
./motion/Pressure.scala:  def *(that: Time) = ??? // returns DynamicViscosity
./space/Volume.scala:  def /(that: Mass) = ??? // returns SpecificVolume (inverse of Density)
[...]

PR #264, opened 2017-06-14
add AreaElectricChargeDensity, ElectricChargeDensity, ElectricChargeM…

underscorenico:
This is the second PR concerning issue #252 .

In this pull request, four new types are added: AreaElectricChargeDensity, ElectricChargeDensity, ElectricChargeMassRatio and LinearElectricChargeDensity.

This new types are needed to implement some unimplemented functions in ElectricCharge.

### DIFF
--- a/shared/src/main/scala/squants/electro/AreaElectricChargeDensity.scala
+++ b/shared/src/main/scala/squants/electro/AreaElectricChargeDensity.scala
@@ -1,0 +1,49 @@
+package squants.electro
+
+import squants.space.Meters
+import squants.{AbstractQuantityNumeric, Area, Dimension, Length, PrimaryUnit, Quantity, SiUnit, UnitConverter, UnitOfMeasure}
+
+/**
+  *
+  * @author Nicolas Vinuesa
+  * @since 1.4
+  *
+  * @param value Double
+  */
+final class AreaElectricChargeDensity private (val value: Double, val unit: AreaElectricChargeDensityUnit)
+  extends Quantity[AreaElectricChargeDensity] {
+
+  def dimension = AreaElectricChargeDensity
+
+  def *(that: Area): ElectricCharge = Coulombs(this.toCoulombsSquareMeters * that.toSquareMeters)
+  def *(that: Length): LinearElectricChargeDensity = CoulombsPerMeter(this.toCoulombsSquareMeters * that.toMeters)
+
+  def toCoulombsSquareMeters = to(CoulombsPerSquareMeter)
+}
+
+object AreaElectricChargeDensity extends Dimension[AreaElectricChargeDensity] {
+  private[electro] def apply[A](n: A, unit: AreaElectricChargeDensityUnit)(implicit num: Numeric[A]) = new AreaElectricChargeDensity(num.toDouble(n), unit)
+  def apply = parse _
+  def name = "AreaElectricChargeDensity"
+  def primaryUnit = CoulombsPerSquareMeter
+  def siUnit = CoulombsPerSquareMeter
+  def units = Set(CoulombsPerSquareMeter)
+}
+
+trait AreaElectricChargeDensityUnit extends UnitOfMeasure[AreaElectricChargeDensity] with UnitConverter {
+  def apply[A](n: A)(implicit num: Numeric[A]) = AreaElectricChargeDensity(n, this)
+}
+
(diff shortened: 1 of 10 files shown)


derekmorr:
Could you be sure to update `ImplicitDimensions` with this?

underscorenico:
Done!

shared/src/main/scala/squants/electro/AreaElectricChargeDensity.scala
```diff
@@ -0,0 +1,49 @@
...
+  def primaryUnit = CoulombsSquareMeters
+  def siUnit = CoulombsSquareMeters
+  def units = Set(CoulombsSquareMeters)
+}
+
+trait AreaElectricChargeDensityUnit extends UnitOfMeasure[AreaElectricChargeDensity] with UnitConverter {
+  def apply[A](n: A)(implicit num: Numeric[A]) = AreaElectricChargeDensity(n, this)
+}
+
+object CoulombsSquareMeters extends AreaElectricChargeDensityUnit with PrimaryUnit with SiUnit {
```
derekmorr: I wonder if this would be better named as `CoulombsPerSquareMeter` ? I don't have a strong opinion about it.

shared/src/main/scala/squants/electro/AreaElectricChargeDensity.scala
```diff
@@ -0,0 +1,49 @@
...
+  def symbol = Coulombs.symbol + "/" + Meters.symbol + "²"
```
derekmorr: should this be a val ?

shared/src/main/scala/squants/electro/ElectricChargeDensity.scala
```diff
@@ -0,0 +1,50 @@
...
+  def primaryUnit = CoulombsCubicMeters
+  def siUnit = CoulombsCubicMeters
+  def units = Set(CoulombsCubicMeters)
+}
+
+trait ElectricChargeDensityUnit extends UnitOfMeasure[ElectricChargeDensity] with UnitConverter {
+  def apply[A](n: A)(implicit num: Numeric[A]) = ElectricChargeDensity(n, this)
+}
+
+object CoulombsCubicMeters extends ElectricChargeDensityUnit with PrimaryUnit with SiUnit {
```
derekmorr: I wonder if this would be better named as CoulombsPerCubicMeter?

shared/src/main/scala/squants/electro/ElectricChargeDensity.scala
```diff
@@ -0,0 +1,50 @@
...
+  def symbol = Coulombs.symbol + "/" + Meters.symbol + "³"
```
derekmorr: Can this be a val?

shared/src/main/scala/squants/electro/LinearElectricChargeDensity.scala
```diff
@@ -0,0 +1,49 @@
...
+  def primaryUnit = CoulombsMeters
+  def siUnit = CoulombsMeters
+  def units = Set(CoulombsMeters)
+}
+
+trait LinearElectricChargeDensityUnit extends UnitOfMeasure[LinearElectricChargeDensity] with UnitConverter {
+  def apply[A](n: A)(implicit num: Numeric[A]) = LinearElectricChargeDensity(n, this)
+}
+
+object CoulombsMeters extends LinearElectricChargeDensityUnit with PrimaryUnit with SiUnit {
```
derekmorr: I wonder if this would be better named as `CoulombsPerMeter` ?

shared/src/main/scala/squants/electro/LinearElectricChargeDensity.scala
```diff
@@ -0,0 +1,49 @@
...
+  def symbol = Coulombs.symbol + "/" + Meters.symbol
```
derekmorr: Can this be a val ?

derekmorr:
This LGTM. I made several minor comments. I'd like some other maintainers to give an opinion on unit names, only because it's hard to change them after we release. Thanks again for doing all of this work!

cquiroz:
I agree with the suggestion of the name changes for units using `Per`

underscorenico:
I completely agree with you. I have found sometimes there is the "per" in the unit names and sometimes there isn't. So maybe if you agree I could pass over every class and standardise it on another PR.

underscorenico:
Corrected unit names, now including "per"

derekmorr:
Thanks again for doing all of this work!

PR #267, opened 2017-06-15
Improve Double and Dimensionless interoperability

### LINKED ISSUE #233: How to use Percent with Money
Use case: add tax percent to net money amount.

Instead of using `BigDecimal`:
```scala
val net = EUR(100)
val tax = BigDecimal(22) / 100
val totalIncludingTaxes: Money  = net + net * tax
```

Is it possible to improve the example using `Percent`?

garyKeorkunian:
Fixes #233

### DIFF
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -6,7 +6,7 @@ addSbtPlugin("org.scala-native" % "sbt-scalajs-crossproject" % "0.1.0")
 
 addSbtPlugin("org.scala-native" % "sbt-scala-native" % "0.2.1")
 
-addSbtPlugin("org.scalariform" % "sbt-scalariform" % "1.6.0")
+addSbtPlugin("org.scalariform" % "sbt-scalariform" % "1.5.1")
 
 addSbtPlugin("com.typesafe.sbt" % "sbt-osgi" % "0.8.0")
 
--- a/shared/src/main/scala/squants/Dimensionless.scala
+++ b/shared/src/main/scala/squants/Dimensionless.scala
@@ -8,7 +8,7 @@
 
 package squants
 
-import squants.time.{ Frequency, Hertz, TimeIntegral }
+import squants.time.{Frequency, Hertz, TimeIntegral}
 
 /**
  * Represents a quantity of some thing for which there is no dimension.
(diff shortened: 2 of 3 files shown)


shared/src/main/scala/squants/package.scala
```diff
@@ -118,4 +119,13 @@ package object squants {
...
   }
+
+  /**
+    * Provides an implicit conversion from Dimensionless to Double, allowing a Dimensionless value
+    * to be used anywhere a Double (or similar primitive) is required
+    *
+    * @param d Dimensionless
+    * @return
+    */
+  implicit def DimensionlessToDouble(d: Dimensionless): Double = d.toEach
```
cquiroz: I think there is a tendency to avoid implicit conversions, preferring something like an implicit class with an explicit `toDouble` method.

On the other hand, I think that won't work on 2.10, right?

garyKeorkunian: To support the feature, we really need to support this conversion.  Not sure how an implicit class would work as no method is being called on the Dimensionless.  It's just being used in an op.  If we don't want to do this, the answer is to force the user to call toEach on the Dimensionless - which is effectively a toDouble (maybe we want to add that as an alias) - in places where Double is required.  In the use case described in the issue we end up with:

```
val net = EUR(100)
val tax = 22.percent
val totalIncludingTax: Money = net + net * tax.toEach (or toDouble)
```

To avoid that explicit call, I think we will need the implicit conversion.

Perhaps having it the package is not a great idea as it will get pulled into scope to readily.  There other implicit conversions like this, for example, timeToScalaDuration and the scalaDurationtoTime methods.  However, they are defined TimeConversion object.

Perhaps it would be best to move this to DimensionalessConversion object.

cquiroz: Right. I don't have a very strong opinion and as you said readability is much better this way

derekmorr: I think we should have the implicit conversion, but not in the main squants package. It's too easy for it to be unknowingly included.

garyKeorkunian: Agreed.  I'll move it to the DimensionlessConversions object.

garyKeorkunian:
The implicit conversion has been moved to the DimensionlessConversions object.

garyKeorkunian:
This is ready to go now

PR #268, opened 2017-06-15
Fix precision issues with Money

### LINKED ISSUE #265: Loss of precision in Money when using Multiplication
While using the Squants library to work with the Money type, I encountered a loss of precision when doing certain calculations. I instantiate the money object by using the Currency object and multiply it with an Integer. In this example: `EUR(9709.06)` as Money and `4992` as Integer.

I tried to instantiate it by using the BigDecimal, as it might be related to the double that we pass to the currency object. (By using `EUR(BigDecimal(9709.06))` or `EUR(BigDecimal("9709.06"))`)
However, this didn't solve the problem, actually I received the same results.

An example to reproduce this:
```
scala> val x: Int = 4992; val y: Money = EUR(9709.06)
x: Int = 4992
y: squants.market.Money = 9709.06 EUR

scala> x * y == y * x
res29: Boolean = true

scala> x * y compareTo y * x
res30: Int = 0

scala> x * y + x * y == (y * x) * 2
res31: Boolean = false
// expected true

scala> x * y + x * y compareTo (y * x) * 2
res32: Int = 1
// expected 0

scala> x * y + x * y amount
warning: there was one feature warning; re-run with -feature for details
res33: BigDecimal = 96935255.039999992

scala> (y * x) * 2 amount
warning: there was one feature warning; re-run with -feature for details
res34: BigDecimal = 96935255.03999999
```

garyKeorkunian:
Also temporarily reverted to Scalariform 1.5.1

Fixes #265

### DIFF
--- a/shared/src/test/scala/squants/market/MoneySpec.scala
+++ b/shared/src/test/scala/squants/market/MoneySpec.scala
@@ -142,7 +142,7 @@ class MoneySpec extends FlatSpec with Matchers {
     x != y should be(right = true)
   }
 
-  it should "return proper result on max/min operation with an implicit MoneyContext in scope" in {
+  it should "return a proper result on max/min operation with an implicit MoneyContext in scope" in {
     val r1 = CurrencyExchangeRate(USD(1), JPY(100))
     val r2 = CurrencyExchangeRate(USD(1), EUR(.75))
     implicit val moneyContext = MoneyContext(USD, defaultCurrencySet, List(r1, r2))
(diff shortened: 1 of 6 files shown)


shared/src/test/scala/squants/market/MoneySpec.scala
```diff
@@ -152,17 +152,39 @@ class MoneySpec extends FlatSpec with Matchers {
...
   }
 
-  it should "return proper result when subtracting like currencies with no MoneyContext in scope" in {
+  it should "return a proper result when subtracting like currencies with no MoneyContext in scope" in {
     USD(5) - USD(2) should be(USD(3))
     USD(5).minus(USD(2)) should be(USD(3))
   }
 
-  it should "return proper result when dividing like currencies with no MoneyContext in scope" in {
+  it should "return a proper result with no additional precision loss when multiplying by a primitive numeric" in {
```
derekmorr: I wonder if we should make this a property based test?

cquiroz: That's a good suggestion in general for test, scala check could be appropriate for that

garyKeorkunian: I'll add a Scalacheck test.

garyKeorkunian: So, the Scalachecks revealed that this issue can also exist with BigDecimal.

```scala
val x = BigDecimal(3)
val m = BigDecimal(2)

m / x + m / x = 1.3333333333333333333333333333333334
(m / x) * 2 = 1.333333333333333333333333333333333
```

Interestingly, that same calculation with Double works correctly

```scala
val x = BigDecimal(3)
val m = BigDecimal(2)

m / x + m / x = 1.3333333333333333
(m / x) * 2 = 1.3333333333333333
```

I can resolve this by using an approximation with a tolerance of USD(1e-32) to the comparison.  Most uses cases will use much less than 32 digits and will typically round to the nearest penny (or formattedDecimal for the Currency), so this is still a big improvement over the 16/17 digits we get with Double math.

The multiplication use case noted in the original issue does not have this problem, so no approximation was needed there.

I'll have an updated PR in a moment.

cquiroz: That's great, we should probably extend scalacheck's usage to other tests

derekmorr:
LGTM. Are we good to merge this?

garyKeorkunian:
Any thoughts on that Checks issue I mentioned?

PR #270, opened 2017-06-19
Adds mass.AreaDensity UOM PoundsPerAcre

tel:
This is a unit we're using at Cibo which wasn't availble. It's easy
enough to define it locally, but since it's valid general code I am
submitting it to Squants itself.

### DIFF
--- a/shared/src/main/scala/squants/mass/AreaDensity.scala
+++ b/shared/src/main/scala/squants/mass/AreaDensity.scala
@@ -26,6 +26,7 @@ final class AreaDensity private (val value: Double, val unit: AreaDensityUnit)
   def toKilogramsPerSquareMeter = to(KilogramsPerSquareMeter)
   def toGramsPerSquareCentimeter = to(GramsPerSquareCentimeter)
   def toKilogramsPerHectare = to(KilogramsPerHectare)
+  def toPoundsPerAcre = to(PoundsPerAcre)
 }
 
 /**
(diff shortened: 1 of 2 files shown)


shared/src/main/scala/squants/mass/AreaDensity.scala
```diff
@@ -59,6 +60,12 @@ object GramsPerSquareCentimeter extends AreaDensityUnit with UnitConverter with
   val conversionFactor = (100*100d)/1000d
 }
 
+object PoundsPerAcre extends AreaDensityUnit with UnitConverter {
+  val symbol = "lbs/acre"
```
derekmorr: Minor nit - could you use the existing lbs and acre symbols? So something like:

```scala
val symbol = s"${Pounds.symbol}/${Acres.symbol}"
```

We're just trying to avoid redefining symbols.

derekmorr:
Hi,

Thanks for the submission. Could you add some tests similar to the other ones in `AreaDentitySpec` ?

-derek

tel:
Yes, revisiting the code on this I'm actually interested in making those changes as well as trying to re-use some of the conversion factors that already exist (instead of just magic-numbering in the one I have).

Something like 

```scala
object PoundsPerAcre extends AreaDensityUnit with UnitConverter {
  val symbol = s"${Pounds.symbol}/${Acres.symbol}"
  // Base unit is kg/m^2
  val conversionFactor = (Pounds.conversionFactor / Kilograms.conversionFactor) / (Acres.conversionFactor / SquareMeters.conversionFactor)
}
```

tel:
My bad, this should fix the tests.

derekmorr:
Thanks again!

Issue #257, opened 2017-06-07
Add physical constants


underscorenico:
An object could be added, containing (some ?) known and widely used physical constants. This constants will be very practical for conversions.
A comprehensive list can be found here: http://physics.nist.gov/cuu/pdf/all_2002.pdf

cquiroz:
I'd be interested on taking this unless somebody else is already working it

garyKeorkunian:
There are a few constants already defined.  The can be found in the package object of the of sub-package that contains that unit.  For example ...

```
package object motion {
   ...
   lazy val SpeedOfLight = Velocity(Meters(299792458e8), Seconds(1))
   ...
}
```

It looks like only the thermo, electro and motion sub-packages currently define a few.

I'm not sure if these package objects are the best place for them.  Perhaps a PhysicalConstants object which contains them all (or one per logical category) makes sense.

Also, the energy sub-package defines a formula for Kinetic Energy.  In addition to the physical constants, perhaps there are other standard formulas we may wish to support.

PR #271, opened 2017-06-20
Permeability

cquiroz:
While looking at the universal constants #257 I realized that we don't have a definition for `Permeability`. This PR adds this new `Quantity`


underscorenico:
Maybe we should standardize what to do with "symbol". In some clases it is a `val`, in others it is a `def`.
@cquiroz @derekmorr

cquiroz:
@underscorenico That's a good observation. I'd venture to say that in most cases a `val` is appropriate. I'll change the code to reflect that

underscorenico:
@cquiroz I was thinking that you could also implement the `/` method in `Inductance`, which returns `Permeability`.

cquiroz:
@underscorenico Good suggestion, I added the division and also added multiplication to `Permeability` to get `Inductance`

cquiroz:
How can we move forward on this and the other pending PRs?

PR #272, opened 2017-06-20
add ElectricFieldStrength, implement ElectricPotential unimplemented method

underscorenico:
This is the 4th PR concerning #252

### DIFF
--- a/shared/src/main/scala/squants/electro/ElectricFieldStrength.scala
+++ b/shared/src/main/scala/squants/electro/ElectricFieldStrength.scala
@@ -1,0 +1,48 @@
+package squants.electro
+
+import squants.space.{Length, Meters}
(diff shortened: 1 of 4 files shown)


shared/src/main/scala/squants/electro/ElectricFieldStrength.scala
```diff
@@ -0,0 +1,48 @@
...
+
+  def *(that: Length): ElectricPotential = Volts(this.toVoltsPerMeter * that.toMeters)
+
+  def toVoltsPerMeter = to(VoltsPerMeter)
+}
+
+object ElectricFieldStrength extends Dimension[ElectricFieldStrength] {
+  private[electro] def apply[A](n: A, unit: ElectricFieldStrengthUnit)(implicit num: Numeric[A]) = new ElectricFieldStrength(num.toDouble(n), unit)
+  def apply = parse _
+  def name = "ElectricFieldStrength"
```
cquiroz: I think some of these could be `val` as we've discussed in other PRs

PR #273, opened 2017-06-23
Energy electron volt

cquiroz:
I noted that we are missing the very important `Electron-Volt` unit for energy. This PR adds them

I'll probably add the other uses of `eV` for `Mass`, `Momentum`,  `Distance` and `Temperature`

Do you think I should add a `UnitGroup` for these?

### DIFF
--- a/shared/src/main/scala/squants/energy/Energy.scala
+++ b/shared/src/main/scala/squants/energy/Energy.scala
@@ -67,6 +67,15 @@ final class Energy private (val value: Double, val unit: EnergyUnit)
   def toGigajoules = to(Gigajoules)
   def toTerajoules = to(Terajoules)
 
+  def toeV = to(ElectronVolt)
+  def tomeV = to(MilliElectronVolt)
+  def tokeV = to(KiloElectronVolt)
+  def toMeV = to(MegaElectronVolt)
+  def toGeV = to(GigaElectronVolt)
+  def toTeV = to(TeraElectronVolt)
+  def toPeV = to(PetaElectronVolt)
+  def toEeV = to(ExaElectronVolt)
+
   def toBtus = to(BritishThermalUnits)
   def toMBtus = to(MBtus)
   def toMMBtus = to(MMBtus)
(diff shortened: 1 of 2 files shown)


derekmorr:
We can't add a unit group for them since they're different Dimensions.

cquiroz:
@derekmorr Wouldn't e.g. `eV`, `keV`, `MeV` belong to the same group?

derekmorr:
Oh, sorry. I misunderstood. I thought you meant adding units for `Mass`, `Momentum`, etc into a unit group.

I agree that naming will be an issue.

shared/src/main/scala/squants/energy/Energy.scala
```diff
@@ -210,6 +261,16 @@ object EnergyConversions {
...
 
+  lazy val eV = ElectronVolt(1)
+  lazy val meV = MilliElectronVolt(1)
+  lazy val keV = KiloElectronVolt(1)
+  lazy val MeV = MegaElectronVolt(1)
+  lazy val GeV = GigaElectronVolt(1)
+  lazy val TeV = TeraElectronVolt(1)
+  lazy val PeV = PetaElectronVolt(1)
+  lazy val EeV = ExaElectronVolt(1)
+  lazy val evMultiplier: Double = 1.60217656535e-19
```
derekmorr: Should this be moved into the `ElectronVolt` object so we don't leak a constant?

cquiroz: Yes, that's probably a good idea

derekmorr:
This LGTM. Did we decide on how to handle naming issues?

cquiroz:
No decision yet, but I think it will be a problem to have the same unit name. I'm leaning to use `ElectronVolt` only for `Energy` and add a prefix in the other cases like `ElectronVoltMass`, `ElectronVoltTemperature`, etc

WDYT?

shared/src/main/scala/squants/energy/Energy.scala
```diff
@@ -187,6 +198,46 @@ object Ergs extends EnergyUnit {
   val symbol = "erg"
 }
 
+object ElectronVolt extends EnergyUnit {
+  val conversionFactor = Joules.conversionFactor * 1.60217656535e-19
```
cquiroz: Thanks for poiting this up, I'll update this and the constants PR that uses the same bad logic

cquiroz:
Seems I could merge this. I'll wait a bit to let others chime in if needed

PR #274, opened 2017-06-23
add MolarEnergy, implement Energy unimplemented method

underscorenico:
This is the 5th PR concerning #252 and the first one implementing unimplemented methods and missing units in energy.

### DIFF
--- a/shared/src/main/scala/squants/energy/MolarEnergy.scala
+++ b/shared/src/main/scala/squants/energy/MolarEnergy.scala
@@ -1,0 +1,48 @@
+package squants.energy
+
+import squants.mass.{ChemicalAmount, Moles}
+import squants.{AbstractQuantityNumeric, Dimension, PrimaryUnit, Quantity, SiUnit, UnitConverter, UnitOfMeasure}
+
+/**
+  *
+  * @author Nicolas Vinuesa
+  * @since 1.4
+  *
+  * @param value Double
+  */
+final class MolarEnergy private (val value: Double, val unit: MolarEnergyUnit)
+  extends Quantity[MolarEnergy] {
+
+  def dimension = MolarEnergy
+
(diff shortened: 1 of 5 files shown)


shared/src/main/scala/squants/energy/MolarEnergy.scala
```diff
@@ -0,0 +1,48 @@
...
+
+import squants.mass.{ChemicalAmount, Moles}
+import squants.{AbstractQuantityNumeric, Dimension, PrimaryUnit, Quantity, SiUnit, UnitConverter, UnitOfMeasure}
+
+/**
+  *
+  * @author Nicolas Vinuesa
+  * @since 1.4
+  *
+  * @param value Double
```
cquiroz: Some description?

cquiroz: I mean, I'm not very familiar with `MolarEnergy`, perhaps we could add a link to wikipedia?

underscorenico: Its just a unit of energy over mass. The link could be: https://en.wikipedia.org/wiki/Joule_per_mole
Should I add it in the class? I've not seen it elsewhere...

cquiroz: Hmm, you're right. I've seen only one case but that is in another level. Nevermind

derekmorr:
LGTM. We'll need to add this to `ImplicitDimensions` too.

underscorenico:
@derekmorr it is already in implicitDimensions ... check the commit

derekmorr:
Ah, sorry. Too many commits this morning.

garyKeorkunian:
I think that misspelling of JoulesToMol*e* should be corrected before this is merged.

underscorenico:
Of course, i'll get it done right away

underscorenico:
@garyKeorkunian done!

PR #275, opened 2017-06-24
Mass electron volt

cquiroz:
This PR adds electron volt as a `Mass` unit.

A possible problem here is that there are `ElectronVolt` units for different quantities and there could be some conflicts if you were going to import both `Energy` and `Mass`. They are different units but use the same name in the common usage


shared/src/main/scala/squants/mass/Mass.scala
```diff
@@ -171,6 +182,45 @@ object SolarMasses extends MassUnit {
   val symbol = "M☉"
 }
 
+object ElectronVolt extends MassUnit {
```
derekmorr: Won't this cause name collisions in the REPL since we import all constants and implicits by default?

cquiroz: This is problematic in general. Having the same name will create problems either on the REPL or other cases. I'm kind of leaning to use `ElectronVolt` for `Energy` and use something like `ElectronVoltMass` in this case and similar

What do you think?

cquiroz:
I renamed the unit to `ElectronVoltMass`. As much as I like the original unit, it is bound to create issues to have the same name for different dimensions

PR #276, opened 2017-06-25
first draft of general tests

derekmorr:
This isn't ready to merge yet, it's just here for discussion.

As discussed in the Gitter channel, there's a lot of copy/paste in the tests. There's also a lot of boilerplate, which makes it tedious to add new Dimensions (something @underscorenico has been doing a lot of lately). I've attempted to extract the common aspects of the tests into a new base class, `GenericSpec`. I made a copy of `LengthSpec` that uses this new base class -- it's in `NewLengthSpec`. We were able to remove over 100 lines of code.

I lean more heavily on ScalaTest for both property-based and table-based tests. I did have to add an `apply()` method to `Dimension` to get the tests to pass. We've gone back and forth on this for a while. Since we've got several external libraries integrating with squants (pureconfig and ciris), adding this would make their lives easier.

I'd like the other maintainers to offer thoughts on this approach before its merged.

### DIFF
--- a/shared/src/main/scala/squants/Dimension.scala
+++ b/shared/src/main/scala/squants/Dimension.scala
@@ -50,12 +50,14 @@ trait Dimension[A <: Quantity[A]] {
    */
   def symbolToUnit(symbol: String): Option[UnitOfMeasure[A]] = units.find(u ⇒ u.symbol == symbol)
 
+  def apply(value: Any): Try[A] = parse(value)
+
   /**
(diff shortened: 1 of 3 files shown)


underscorenico:
I completely agree with you, when adding new dimensions its very easy to get tempted with copy/pasting tests. This should improve our code quality a lot

cquiroz:
I think this is great. There is a lot of repetition on the tests as you mention

PR #277, opened 2017-06-26
Upgrade Scala Native to 0.3.0

### LINKED ISSUE #266: upgrade to scala native 0.3
Scala Native 0.3 is out - https://github.com/scala-native/scala-native/releases/tag/v0.3.0. We should upgrade to it and see if we can remove some of the work-arounds we had to implement in the past.

shadaj:
This resolves #266, but still has tests disabled because ScalaTest does not support Scala Native yet.


derekmorr:
Ok. Sounds like we should go ahead and merge this now. We can enable tests later once ScalaTest is ready.

Thanks for doing this!

PR #278, opened 2017-06-29
RFC: Universal constants

cquiroz:
This is a request for comments for implementing `Constants` on squants. Instead of just adding named `Quantity` values I thought about adding a `Constant` type that would have some more information about the constant, including its symbol and uncertainty

I implemented a few examples of `Universal` and `ElectroMagnetic` constants but there are many more to come

Some aspects to discuss:

* Do you think the encoding is sound?
* Some constants have more that one representation in a different set of units, Should we accomodate that?
* Probably the biggest issue is that some constants (Like the Gravitational Constants) aren't really a known quantity, instead they have units that are only pertinent to that constant (In the case of G it is m3 kg-1 s-2). I'm not too keen on just having a quantity for that particular use but I see no other options. What do you think


derekmorr:
LGTM.

underscorenico:
LGTM

golem131:
@cquiroz maybe it would be better to start from definition of base SI Units (https://en.wikipedia.org/wiki/Proposed_redefinition_of_SI_base_units), such as:

- Speed of light (already here)
- Planck constant
- elementary charge (e)
- Boltzmann constant
- Avogadro constant 
- ground state hyperfine splitting frequency of the caesium-133 atom 
- luminous efficacy

cquiroz:
I'm wondering how to go from here. My main contention point is how to represent constants without an actual representation in a given unit. Like the gravitational constant in m3 kg-1 s-2

garyKeorkunian:
I like adding the uncertainty value.  I think we could do something with the approx operators but I haven't thought about specific use cases.

As for the Gravitational Constant, I agree that's a tough one.  We do have something called TimeSquared, which is not an actual Quantity, but an intermediate value that can be composed with other quantities to make valid quantities.  Perhaps we could do something like that if it is useful.  

Another possibility is to create a value for the Gravitational Coupling Constant, which is dimensionless. I don't know enough about how these values are used to know for sure.

PR #279, opened 2017-07-13
Electron volt as a length unit

cquiroz:
Among may uses electron volt can be used for distance too


underscorenico:
Is electronVoltLength a real unit? I can't find it anywhere...

cquiroz:
https://en.wikipedia.org/wiki/Electronvolt#Distance

underscorenico:
oh great!

PR #282, opened 2017-08-09
Bump version to Scala Native 0.3.2 and make all unit tests pass

shadaj:
With the PR at https://github.com/scalatest/scalatest/pull/1185, all unit tests now pass in Scala Native!

### DIFF
--- a/shared/src/main/scala/squants/Dimension.scala
+++ b/shared/src/main/scala/squants/Dimension.scala
@@ -97,10 +97,8 @@ trait Dimension[A <: Quantity[A]] {
     }
   }
 
-  private lazy val regexUnits = units
-
   private lazy val QuantityString =
-    ("^([-+]?[0-9]*\\.?[0-9]+(?:[eE][-+]?[0-9]+)?) *(" + regexUnits.map { u: UnitOfMeasure[A] ⇒ u.symbol }.reduceLeft(_ + "|" + _) + ")$").r
+    ("^([-+]?[0-9]*\\.?[0-9]+(?:[eE][-+]?[0-9]+)?) *(" + units.map { u: UnitOfMeasure[A] ⇒ u.symbol }.reduceLeft(_ + "|" + _) + ")$").r
 
   private def parseTuple(value: Double, symbol: String): Try[A] = {
     symbolToUnit(symbol) match {
--- a/shared/src/main/scala/squants/Dimension.scala
+++ b/shared/src/main/scala/squants/Dimension.scala
@@ -55,8 +55,32 @@ trait Dimension[A <: Quantity[A]] {
    * @param value the source string (ie, "10 kW") or tuple (ie, (10, "kW"))
    * @return Try[A]
    */
-  protected def parse(value: Any) = value match {
-    case s: String              => parseString(s)
+  protected def parse(value: Any): Try[A] = value match {
+    case s: String              =>
+      if (Platform.name == "native") {
+        // inefficient but correct parsing
+        units.flatMap { uom =>
+          if (s.endsWith(uom.symbol)) {
+            val remaining = s.dropRight(uom.symbol.length).trim
+            val doubleRegex = "\\d+(.\\d+)?".r
+
+            remaining match {
+              case doubleRegex(_) =>
+                Some(uom(remaining.toDouble))
+              case _ =>
+                None
+            }
+          } else {
+            None
+          }
+        }.headOption.map { v =>
+          Success(v)
+        }.getOrElse {
+          Failure(QuantityParseException(s"Unable to parse $name", s))
+        }
+      } else {
+        parseString(s)
+      }
     case (v: Byte, u: String)   => parseTuple(v.toDouble, u)
     case (v: Short, u: String)  => parseTuple(v.toDouble, u)
     case (v: Int, u: String)    => parseTuple(v.toDouble, u)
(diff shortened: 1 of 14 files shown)


shared/src/main/scala/squants/Dimension.scala
```diff
@@ -72,7 +96,11 @@ trait Dimension[A <: Quantity[A]] {
       case _                             ⇒ Failure(QuantityParseException(s"Unable to parse $name", s))
     }
   }
-  private lazy val QuantityString = ("^([-+]?[0-9]*\\.?[0-9]+(?:[eE][-+]?[0-9]+)?) *(" + units.map { u: UnitOfMeasure[A] ⇒ u.symbol }.reduceLeft(_ + "|" + _) + ")$").r
+
+  private lazy val regexUnits = units
```
derekmorr: Why was this alias made?

shared/src/main/scala/squants/Dimension.scala
```diff
@@ -55,8 +55,32 @@ trait Dimension[A <: Quantity[A]] {
    * @param value the source string (ie, "10 kW") or tuple (ie, (10, "kW"))
    * @return Try[A]
    */
-  protected def parse(value: Any) = value match {
-    case s: String              => parseString(s)
+  protected def parse(value: Any): Try[A] = value match {
```
derekmorr: Why was this change necessary?

shared/src/test/scala/squants/thermal/TemperatureSpec.scala
```diff
@@ -288,7 +288,12 @@ class TemperatureSpec extends FlatSpec
     c.toString(Kelvin) should be("273.15°K")
     c.toString(Fahrenheit) should be("32.0°F")
     c.toString(Celsius) should be("0.0°C")
-    c.toString(Rankine) should be("491.66999999999996°R")
+
+    if (squants.Platform.name == "native") {
```
derekmorr: Why are these native-specific changes necessary?

shared/src/main/scala/squants/thermal/Temperature.scala
```diff
@@ -141,14 +141,24 @@ object Temperature extends Dimension[Temperature] with BaseDimension {
...
-    val regex = "([-+]?[0-9]*\\.?[0-9]+(?:[eE][-+]?[0-9]+)?) *°? *(f|F|c|C|k|K|r|R)".r
-    s match {
-      case regex(value, unit) => unit match {
-        case "f" | "F" => Success(Fahrenheit(value.toDouble))
-        case "c" | "C" => Success(Celsius(value.toDouble))
-        case "k" | "K" => Success(Kelvin(value.toDouble))
-        case "r" | "R" => Success(Rankine(value.toDouble))
-      }
+    val regex = if (squants.Platform.name == "native") {
+      "([-+]?[0-9]*\\.?[0-9]+(?:[eE][-+]?[0-9]+)?) *(__DEGREE__)? *(f|F|c|C|k|K|r|R)".r
```
derekmorr: I'm assuming the native platform doesn't support the unicode degree symbol, which is what prompted this change. I'm wondering if we can factor out just that into a conditional to avoid copy/pasting the regex.

Something like:
```scala
val degSymbol = if (squants.Platform.name == "native") "__DEGREE__" else "°"
val regex = s"([-+]?[0-9]*\\.?[0-9]+(?:[eE][-+]?[0-9]+)?) *${degSymbol}? *(f|F|c|C|k|K|r|R)".r
```

shared/src/main/scala/squants/package.scala
```diff
@@ -75,6 +75,33 @@ package object squants {
   type Price[A <: Quantity[A]] = squants.market.Price[A]
 
   /**
+    * Helper function to achieve uniform Double formatting over JVM and JS platforms.
+    * Simple Double.toString will format 1.0 as "1.0" on JVM and as "1" on JS
+    * @param d Double number to be formatted
+    * @return
+    */
+  private[squants] def crossFormat(d: Double): String = {
```
derekmorr: This version of `crossFormat` isn't the same as before. Did something change in scala native that required us to change?

shared/src/main/scala/squants/Dimension.scala
shadaj: Scala Native can't handle regexes with unicode yet, so I added an alternate parsing method for native only.

shared/src/main/scala/squants/Dimension.scala
shadaj: Oops, that was left behind from figuring out how to handle unicode, I'll switch it back to how it was before.

shared/src/main/scala/squants/package.scala
shadaj: Scala Native now supports formatting, so that can now be shared across platforms, but with the toString implementation on Double right now Scala Native includes trailing zeroes, which would break a lot of unit tests.

shared/src/main/scala/squants/thermal/Temperature.scala
shadaj: That sounds good, I'll make that change.

shared/src/test/scala/squants/thermal/TemperatureSpec.scala
shadaj: With Scala Native, the Double toString ends up rounding to "491.67", so until that behavior is fixed I adjusted the tests.

shadaj:
Looks like we won't need the patches for regexes soon, with https://github.com/scala-native/scala-native/pull/946/files on its way!

derekmorr:
Once the tests pass for this PR, we should bump the Scala Native version to 0.3.3, which was just release today.

shadaj:
@derekmorr I'm going to close this PR and work on upgrading directly to Scala Native 0.3.3, since that fixes many of the issues with regexes

PR #285, opened 2017-10-24
FIX #281

### LINKED ISSUE #281: Missing function when using multiplication with the Money type
For my project I am automatically checking some properties on the different types used in a system, and an error occured when using the Money class. The following error occured when checking the associative property: `(x * y) * z == x * (y * z)`. 
```
<console>:18: error: overloaded method value * with alternatives:
  (x: Double)Double <and>
  (x: Float)Float <and>
  (x: Long)Long <and>
  (x: Int)Int <and>
  (x: Char)Int <and>
  (x: Short)Int <and>
  (x: Byte)Int
 cannot be applied to (squants.market.Money)
       x * (y * z)
         ^
```
I have checked it with both the 1.3 version and the latest snapshot, but unfortunately this didn't solve it.

The error occurs on the right hand side of the equation and is probably happening because of the **parenthesis**.  The types do matter here, as other definitions that I have relating to associativity are working as expected. In this case: 
- x: Int
- y: Int
- z: Money

The following steps can be taken to reproduce this:
```
scala> val x: Int = 2; val y: Int = 4; val z: Money = EUR(50)
x: Int = 2
y: Int = 4
z: squants.market.Money = 5E+1 EUR

scala> x * y * z // Works as expected
res0: squants.market.Money = 4E+2 EUR

scala> (x * y) * z // Works as expected
res1: squants.market.Money = 4E+2 EUR

scala> x * (y * z) // Error
<console>:18: error: overloaded method value * with alternatives:
  (x: Double)Double <and>
  (x: Float)Float <and>
  (x: Long)Long <and>
  (x: Int)Int <and>
  (x: Char)Int <and>
  (x: Short)Int <and>
  (x: Byte)Int
 cannot be applied to (squants.market.Money)
       x * (y * z) // Error
         ^
scala> (z * y) * x // Switching 'x' to the right side works though
res3: squants.market.Money = 4E+2 EUR

scala>
```

golem131:
Fix #281

### DIFF
--- a/shared/src/main/scala/squants/package.scala
+++ b/shared/src/main/scala/squants/package.scala
@@ -105,6 +105,21 @@ package object squants {
   }
 
   /**
+    * Provides implicit conversions that allow Int to lead in * and / by Time operations
+    * {{{
+    *    5 * Kilometers(10) should be(Kilometers(15))
+    * }}}
+    *
+    * @param l Int
+    */
+  implicit class SquantifiedInt(l: Int) {
+    def *[A <: Quantity[A]](that: A): A = that * l.toDouble
+    def *[A](that: SVector[A]): SVector[A] = that * l.toDouble
+    def /(that: Time) = Each(l) / that
+    def per(that: Time): Frequency = /(that)
+  }
+
+  /**
    * Provides implicit conversions that allow BigDecimals to lead in * and / by Time operations
    * {{{
    *    BigDecimal(1.5) * Kilometers(10) should be(Kilometers(15))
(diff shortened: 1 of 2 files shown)


shared/src/main/scala/squants/package.scala
```diff
@@ -105,6 +105,21 @@ package object squants {
   }
 
   /**
+    * Provides implicit conversions that allow Int to lead in * and / by Time operations
+    * {{{
+    *    5 * Kilometers(10) should be(Kilometers(15))
+    * }}}
+    *
+    * @param l Long
```
derekmorr: we should change the doc from `long` to `int` since that's what it really is.

golem131:
@derekmorr fixed

PR #287, opened 2017-11-29
Fix inappropriate multiplication in conversion factors

psung:
Many `conversionFactor`s were expressed in a way similar to the following

```
object FeetPerSecondSquared extends AccelerationUnit {
  [...]
  val conversionFactor = Feet.conversionFactor * Meters.conversionFactor
}
```

The `conversionFactor` should in principle be unitless (giving as a `Double`
the number of meters, or whatever the base unit is, in a foot) so I believe
they should have been written as:

```
object FeetPerSecondSquared extends AccelerationUnit {
  [...]
  val conversionFactor = Feet.conversionFactor / Meters.conversionFactor
}
```

Probably why this was never noticed was because `Meters.conversionFactor` has
the value 1.0, so it doesn't matter whether you multiply or divide by it. While
the existing code is functionally correct, I fixed all instances of this to
make it more clear how unit conversions should in general be performed.

As a sanity-check, I verified that all the units whose conversion factor I
moved into the denominator have a `conversionFactor` of 1.0: Radians, Meters,
Newtons, and CubicMeters.


derekmorr:
Thanks for the PR. Sorry this took so long to get approved and merged.

PR #290, opened 2017-12-19
Add units of measure: mm/s (velocity), mm/s^2 (acceleration), ng (mass)

psung:
These units give the next smaller SI prefix than what is currently available for velocity, acceleration, and mass.


psung:
FYI, we have uses for all of these units internally, but I don't know what `squants`'s general policy is on having derived units that only differ in an SI prefix from some existing unit. For example, I'd completely understand if you decide you don't want to have to pave the way, for, say, having every unit of acceleration from `NanometersPerSecondSquared` to `KilometersPerSecondSquared`, or whatever. So feel free to only accept all/some/none of these changes; in the branch I've separated each of the new units into a separate commit.

derekmorr:
This looks good to me. Thanks for adding these.

PR #292, opened 2018-01-17
Make Dimension.parse public

garyKeorkunian:
This change is at the request of developers of companion libraries.

Making `Dimension.parse` public enable these libraries to integrate more abstractly, preventing the need to code for each Quantity type.


garyKeorkunian:
Closing, to refactor parse(Any)

PR #293, opened 2018-01-18
Make Dimension.parseString/Tuple public

garyKeorkunian:
This change is at the request of developers of companion libraries.

Making `Dimension.parseString` and `Dimension.parseTuple` public to enable these libraries to integrate more abstractly, preventing the need to code for each Quantity type.

This change also updates `parseTuple` to use any Numeric instead of just Double.

The signature of the `parse` method proper is being left as is and protected.  Refactoring and exposing this method will break existing Dimension implementations and is, therefore, being deferred for now.

### DIFF
--- a/shared/src/main/scala/squants/Dimension.scala
+++ b/shared/src/main/scala/squants/Dimension.scala
@@ -56,13 +56,14 @@ trait Dimension[A <: Quantity[A]] {
    * @return Try[A]
    */
   protected def parse(value: Any): Try[A] = value match {
-    case s: String              => parseString(s)
-    case (v: Byte, u: String)   => parseTuple((v, u))
-    case (v: Short, u: String)  => parseTuple((v, u))
-    case (v: Int, u: String)    => parseTuple((v, u))
-    case (v: Long, u: String)   => parseTuple((v, u))
-    case (v: Float, u: String)  => parseTuple((v, u))
-    case (v: Double, u: String) => parseTuple((v, u))
+    case s: String              ⇒ parseString(s)
+    case (v: Byte, u: String)   ⇒ parseTuple((v, u))
+    case (v: Short, u: String)  ⇒ parseTuple((v, u))
+    case (v: Int, u: String)    ⇒ parseTuple((v, u))
+    case (v: Long, u: String)   ⇒ parseTuple((v, u))
+    case (v: Float, u: String)  ⇒ parseTuple((v, u))
+    case (v: Double, u: String) ⇒ parseTuple((v, u))
+    case _ ⇒ Failure(QuantityParseException(s"Unable to parse $name", value.toString))
   }
 
   def parseString(s: String): Try[A] = {
@@ -73,15 +74,15 @@ trait Dimension[A <: Quantity[A]] {
   }
   private lazy val QuantityString = ("^([-+]?[0-9]*\\.?[0-9]+(?:[eE][-+]?[0-9]+)?) *(" + units.map { u: UnitOfMeasure[A] ⇒ u.symbol }.reduceLeft(_ + "|" + _) + ")$").r
 
-  def parseTuple[N: Numeric](t: (N, String)): Try[A] = {
+  def parseTuple[N](t: (N, String))(implicit num: Numeric[N]): Try[A] = {
     symbolToUnit(t._2) match {
       case Some(unit) ⇒ Success(unit(t._1))
-      case None       ⇒ Failure(QuantityParseException(s"Unable to identify $name unit ${t._2}", (t._1, t._2).toString()))
+      case None       ⇒ Failure(QuantityParseException(s"Unable to identify $name unit ${t._2}", s"(${Platform.crossFormat(num.toDouble(t._1))},${t._2})"))
     }
   }
 }
 
-case class QuantityParseException(message: String, expression: String) extends Exception
+case class QuantityParseException(message: String, expression: String) extends Exception(message + ":" + expression)
 
 /**
  * SI Base Quantity
--- a/shared/src/main/scala/squants/Dimension.scala
+++ b/shared/src/main/scala/squants/Dimension.scala
@@ -55,18 +55,17 @@ trait Dimension[A <: Quantity[A]] {
    * @param value the source string (ie, "10 kW") or tuple (ie, (10, "kW"))
    * @return Try[A]
    */
-  protected def parse(value: Any) = value match {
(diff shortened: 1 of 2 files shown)


shared/src/main/scala/squants/Dimension.scala
```diff
@@ -55,29 +55,28 @@ trait Dimension[A <: Quantity[A]] {
...
-    case (v: Int, u: String)    => parseTuple(v.toDouble, u)
-    case (v: Long, u: String)   => parseTuple(v.toDouble, u)
-    case (v: Float, u: String)  => parseTuple(v.toDouble, u)
-    case (v: Double, u: String) => parseTuple(v, u)
+    case (v: Byte, u: String)   => parseTuple((v, u))
+    case (v: Short, u: String)  => parseTuple((v, u))
+    case (v: Int, u: String)    => parseTuple((v, u))
+    case (v: Long, u: String)   => parseTuple((v, u))
+    case (v: Float, u: String)  => parseTuple((v, u))
+    case (v: Double, u: String) => parseTuple((v, u))
```
derekmorr: should we add a default case that returns a failure ?

shared/src/main/scala/squants/Dimension.scala
```diff
@@ -55,29 +55,28 @@ trait Dimension[A <: Quantity[A]] {
...
       case _                             ⇒ Failure(QuantityParseException(s"Unable to parse $name", s))
     }
   }
   private lazy val QuantityString = ("^([-+]?[0-9]*\\.?[0-9]+(?:[eE][-+]?[0-9]+)?) *(" + units.map { u: UnitOfMeasure[A] ⇒ u.symbol }.reduceLeft(_ + "|" + _) + ")$").r
 
-  private def parseTuple(value: Double, symbol: String): Try[A] = {
-    symbolToUnit(symbol) match {
-      case Some(unit) ⇒ Success(unit(value))
-      case None       ⇒ Failure(QuantityParseException(s"Unable to identify $name unit $symbol", (value, symbol).toString()))
+  def parseTuple[N: Numeric](t: (N, String)): Try[A] = {
```
derekmorr: I think we should use the alternate syntax for the constraint:

```scala
def parseTuple[N](value: N, symbol: String)(implicit num: Numeric[N]): Try[A]
```

I believe this syntax is preferred by the Scala style guide.

shared/src/test/scala/squants/QuantitySpec.scala
```diff
@@ -605,4 +606,81 @@ class QuantitySpec extends FlatSpec with Matchers with CustomMatchers {
...
+    implicit val time = Time
+    implicit val thingee = Thingee
+    implicit val mass = Mass
+
+    val l = parse[Length]((100d, "m"))
+    val t = parse[Time]((100d, "m"))
+    val th = parse[Thingee]((100d, "th"))
+    val m = parse[Mass]((100d, "m"))
+
+    l.get should be(Meters(100d))
```
derekmorr: rather than calling `.get()` on the `Try`s, I think the code would look cleaner with ScalaTest's `TryValues` - http://doc.scalatest.org/2.2.6/#org.scalatest.TryValues.

There's an example (using `EitherValues`) in the docs - http://www.scalatest.org/user_guide/other_goodies#eitherValues. It's straightforward to convert to `TryValues`.

garyKeorkunian:
I'll make the changes you suggest and update the PR soon.  Thanks for your review.  Great comments!

garyKeorkunian:
@derekmorr All review comments have been addressed.  The TryValue approach also revealed a bug with QuantityParseException - it was extending Exception with passing through the message, so it was always null.  This has been fixed.  Thanks again.

shared/src/main/scala/squants/Dimension.scala
```diff
@@ -55,34 +55,34 @@ trait Dimension[A <: Quantity[A]] {
...
+  def parseTuple[N](t: (N, String))(implicit num: Numeric[N]): Try[A] = {
+    symbolToUnit(t._2) match {
+      case Some(unit) ⇒ Success(unit(t._1))
+      case None       ⇒ Failure(QuantityParseException(s"Unable to identify $name unit ${t._2}", s"(${Platform.crossFormat(num.toDouble(t._1))},${t._2})"))
     }
   }
 }
 
-case class QuantityParseException(message: String, expression: String) extends Exception
+case class QuantityParseException(message: String, expression: String) extends Exception(message + ":" + expression)
```
cquiroz: Perhaps the string here should be built using string interpolation?

shared/src/main/scala/squants/Dimension.scala
```diff
@@ -55,34 +55,34 @@ trait Dimension[A <: Quantity[A]] {
...
    */
-  protected def parse(value: Any) = value match {
-    case s: String              => parseString(s)
-    case (v: Byte, u: String)   => parseTuple(v.toDouble, u)
-    case (v: Short, u: String)  => parseTuple(v.toDouble, u)
-    case (v: Int, u: String)    => parseTuple(v.toDouble, u)
-    case (v: Long, u: String)   => parseTuple(v.toDouble, u)
-    case (v: Float, u: String)  => parseTuple(v.toDouble, u)
-    case (v: Double, u: String) => parseTuple(v, u)
+  protected def parse(value: Any): Try[A] = value match {
```
cquiroz: Nice to get the return type explicitly stated

shared/src/main/scala/squants/Dimension.scala
```diff
@@ -55,34 +55,34 @@ trait Dimension[A <: Quantity[A]] {
...
     }
   }
   private lazy val QuantityString = ("^([-+]?[0-9]*\\.?[0-9]+(?:[eE][-+]?[0-9]+)?) *(" + units.map { u: UnitOfMeasure[A] ⇒ u.symbol }.reduceLeft(_ + "|" + _) + ")$").r
 
-  private def parseTuple(value: Double, symbol: String): Try[A] = {
-    symbolToUnit(symbol) match {
-      case Some(unit) ⇒ Success(unit(value))
-      case None       ⇒ Failure(QuantityParseException(s"Unable to identify $name unit $symbol", (value, symbol).toString()))
+  def parseTuple[N](t: (N, String))(implicit num: Numeric[N]): Try[A] = {
+    symbolToUnit(t._2) match {
```
derekmorr: this is nitpicking, but can we add

```scala
val num: t._1
val unit: t._2
```
here? I think that would make line 80 easier to read.

PR #294, opened 2018-01-22
Add currencies ZAR and NAD.  Add MoneyContext.withAdditionalCurrencies

### LINKED ISSUE #291: Add support for ZAR and NAD currencies
We use `squants` in our project to handle money and currencies, and we need to handle some values in `ZAR` (South Australian Rand) and `NAD` (Namibian Dollar). 

We added the two currencies to our model as follows:

```scala
object ZAR extends Currency("ZAR", "South African Rand", "R", 2)
object NAD extends Currency("NAD", "Namibian Dollar", "N$", 2)
```

We would need to add them also to the `defaultCurrencySet`:
https://github.com/typelevel/squants/blob/81cfd3d5271bafc3ccf57c9437c4c0ca4cedb206/shared/src/main/scala/squants/market/package.scala#L58
so that the following constructor works:
https://github.com/typelevel/squants/blob/720a6ba692f09b026a93bd42ad7a0557b06e23fe/shared/src/main/scala/squants/market/Money.scala#L390

Otherwise, when instantiating `Money` with the string `"ZAR"` as second argument we get an exception:

image
(we receive string values from an external API, so we need to use such constructor)

It would be nice to have all this directly integrated in squants by adding support for the two currencies.

garyKeorkunian:
Fixes #291

### DIFF
--- a/shared/src/main/scala/squants/market/Money.scala
+++ b/shared/src/main/scala/squants/market/Money.scala
@@ -450,6 +450,8 @@ object XAU extends Currency("XAU", "Gold", "oz", 4)
 object BTC extends Currency("BTC", "BitCoin", "B", 15)
 object ETH extends Currency("ETH", "Ether", "\u039e", 15)
 object LTC extends Currency("LTC", "Litecoin", "\u0141", 15)
+object ZAR extends Currency("ZAR", "South African Rand", "R", 2)
+object NAD extends Currency("NAD", "Namibian Dollar", "N$", 2)
 
 /**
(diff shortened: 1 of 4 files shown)


derekmorr:
I'd love to find a good way to solve this in a general way, since we have the same problem with every other Dimension. For now, `withAdditionalCurrencies` is workable.

garyKeorkunian:
Yes, it would be nice to get that into the dimensions.  Of course, MarketContext is a class so it's a bit easier here.

PR #302, opened 2018-04-19
Migrations from Dotty

allanrenucci:
This changes were needed to make squants compile with Dotty (as part of our community build). I though I might as well upstream them


cquiroz:
Thanks for your PR. It seems it covers various changes:
* added tut plugin
* update to sbt 1
* changes to the code (I presume for dotty support)

Is this correct? Perhaps we should split it in several PRs?
is it possible to run the tests in dotty?

cquiroz:
Thanks for clarifying the scope of the plugin, it would be nice to add it to the Travis build but I guess it could be done in another PR

shared/src/main/scala/squants/experimental/unitgroups/si/strict/package.scala
```diff
@@ -13,7 +13,7 @@ package object strict {
   object implicits {
     implicit def mkSiUnitGroup[A <: Quantity[A]](implicit dimension: Dimension[A]): UnitGroup[A] = {
       new UnitGroup[A] {
-        val units: Set[UnitOfMeasure[A]] = dimension.units.collect { case si: SiUnit => si }
+        val units: Set[UnitOfMeasure[A]] = dimension.units.filter(_.isInstanceOf[SiUnit])
```
cquiroz: Why is this change required? I'd rather not have `isInstanceOf` in the code base

shared/src/main/scala/squants/Dimensionless.scala
```diff
@@ -47,7 +47,7 @@ final class Dimensionless private (val value: Double, val unit: DimensionlessUni
  */
 object Dimensionless extends Dimension[Dimensionless] {
   def apply[A](n: A, unit: DimensionlessUnit)(implicit num: Numeric[A]) = new Dimensionless(num.toDouble(n), unit)
-  def apply = parse _
+  def apply(value: Any) = parse(value)
```
cquiroz: Is this change required by dotty?

shared/src/main/scala/squants/experimental/unitgroups/si/strict/package.scala
allanrenucci: This a bug in Dotty. See lampepfl/dotty#3208. It is not clear if this will be fixed as one can always rewrite the code as:
```scala
dimension.units.collect { case si: UnitOfMeasure[A] with SiUnit => si }
```
However, I though using `collect` here was overkill since what it really doe is `filter`

shared/src/main/scala/squants/Dimensionless.scala
allanrenucci: Yes. See this [comment](https://github.com/lampepfl/dotty/issues/2994#issuecomment-326533372)

cquiroz: 👍

shared/src/main/scala/squants/experimental/unitgroups/si/strict/package.scala
cquiroz: I'd rather have the `collect` call than `isInstanceof`

cquiroz:
Are there any other opinions about this PR? Otherwise, I could merge it

PR #303, opened 2018-04-25
Added very large units of length: kpc, Mpc, Gpc

robotsnowfall:
These units are handy when expressing very large distances, such as the distance between stars or galaxies.

See [Parsecs_and_kiloparsecs](https://en.wikipedia.org/wiki/Parsec#Parsecs_and_kiloparsecs) for use cases.


robotsnowfall:
Anything else I need to do here?

cquiroz:
Since there's been no comments I think this is safe to merge

PR #305, opened 2018-05-17
Use money context currencies when parsing a string

estsauver:
This patch adds support for parsing strings including additional
currencies from the moneycontext.

```scala
  import squants.market.{Currency, Money, MoneyContext}
  object KES extends Currency("KES", "Kenyan Shillings", "KES", 0)
  implicit val moneyContext: MoneyContext = MoneyContext(KES, defaultCurrencySet, Nil)

  Money("100 KES")
```

### DIFF
--- a/shared/src/main/scala/squants/market/Money.scala
+++ b/shared/src/main/scala/squants/market/Money.scala
@@ -392,10 +392,10 @@ object Money extends Dimension[Money] {
   def apply[A](n: A, currency: Currency)(implicit num: Numeric[A]) = new Money(BigDecimal(num.toDouble(n)))(currency)
   def apply[A](n: A, currency: String)(implicit num: Numeric[A]) = new Money(BigDecimal(num.toDouble(n)))(defaultCurrencyMap(currency))
 
-  def apply(s: String): Try[Money] = {
-    lazy val regex = ("([-+]?[0-9]*\\.?[0-9]+) *(" + defaultCurrencySet.map(_.code).reduceLeft(_ + "|" + _) + ")").r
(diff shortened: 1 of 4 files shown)


shared/src/main/scala/squants/market/Money.scala
```diff
@@ -392,10 +392,10 @@ object Money extends Dimension[Money] {
   def apply[A](n: A, currency: Currency)(implicit num: Numeric[A]) = new Money(BigDecimal(num.toDouble(n)))(currency)
   def apply[A](n: A, currency: String)(implicit num: Numeric[A]) = new Money(BigDecimal(num.toDouble(n)))(defaultCurrencyMap(currency))
 
-  def apply(s: String): Try[Money] = {
-    lazy val regex = ("([-+]?[0-9]*\\.?[0-9]+) *(" + defaultCurrencySet.map(_.code).reduceLeft(_ + "|" + _) + ")").r
+  def apply(s: String)(implicit moneyContext: MoneyContext): Try[Money] = {
```
cquiroz: I'm not sure about this, passing this as an `implicit` sounds like an overstretch of `implicit`. Wouldn't you rather pass it as a plain param?

estsauver:
It looks like money context is being used implicitly throughout the apis
for money operations. I think the expected use case is to always have a
money context available when working with money.

garyKeorkunian:
I believe this has been corrected by #330

cquiroz:
@garyKeorkunian Do you think this can be closed then?

garyKeorkunian:
@cquiroz Yes, I think so.

PR #307, opened 2018-05-25
Fix Quantity hashcode method.

Passarinho4:
Hi, 
I've fixed hashcode method in Quantity class. The previous one was breaking the equals/hashcode conduct. It is a very serious bug, so please review it and merge if only possible.

### DIFF
--- a/shared/src/main/scala/squants/Quantity.scala
+++ b/shared/src/main/scala/squants/Quantity.scala
@@ -8,6 +8,8 @@
 
 package squants
 
+import java.util.Objects
+
 import scala.math.BigDecimal.RoundingMode
 import scala.math.BigDecimal.RoundingMode.RoundingMode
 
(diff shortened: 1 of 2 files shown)


cquiroz:
Would you mind adding tests. This is serious enough to warrant some automated test

Passarinho4:
I've added simple hashcode test. Is this enough? :)

Passarinho4:
Anyone? @garyKeorkunian @cquiroz @nornagon 
I believe that this is very crucial fix and should be merged and released asap. Is this project alive?

cquiroz:
That looks good for me

Passarinho4:
Ok, so could you ack the review, because I cannot merge it without your permission? And what about releasing the new version?

derekmorr:
Thanks for the PR. Sorry for the delay.

Passarinho4:
Whenever hashcode is invoked on the same object it should return the same value. This is the reason for this test. Thanks! :)

shared/src/main/scala/squants/Quantity.scala
```diff
@@ -176,7 +178,9 @@ abstract class Quantity[A <: Quantity[A]] extends Serializable with Ordered[A] {
    *
    * @return
    */
-  override def hashCode() = toString.hashCode()
+  override def hashCode() = {
+    Objects.hash(dimension, Double.box(to(dimension.primaryUnit)))
```
julienrf, line 182: Unfortunately, the resulting hashCode is not stable across JVM runs because the `dimension` itself may not have a stable `hashCode`.

PR #313, opened 2018-09-03
Inches of Mercury unit of Pressure

fvalka:
I added the unit in Pressure and extended the unit tests. 

The conversion factor is as published in: 
Barry N. Taylor, Guide for the Use of the International System of Units (SI), 1995, 
NIST Special Publication 811, Appendix B
Online: https://www.nist.gov/physical-measurement-laboratory/nist-guide-si-appendix-b8

This pressure unit is commonly used in aviation, for example: In the altimeter setting and in weather reports called METARs.


shared/src/main/scala/squants/motion/Pressure.scala
```diff
@@ -35,6 +35,7 @@ final class Pressure private (val value: Double, val unit: PressureUnit)
   def toPoundsPerSquareInch: Double  = to(PoundsPerSquareInch)
   def toStandardAtmospheres: Double  = to(StandardAtmospheres)
   def toMillimetersOfMercury: Double = to(MillimetersOfMercury)
+  def toInchesOfMercury: Double       = to(InchesOfMercury)
```
cquiroz: Could you align on `=`?

Issue #304, opened 2018-05-08
Add support for {Gibi|Mebi|Kilo|etc}bits


danielyli:
The `squants.information` package currently supports units like `MiB/s` but not bit-based units like `Mbps` (megabits per second), which is the standard unit for telecommunications.

If Squants is to support these, we should consider differentiating between units like “megabits” (10^6 bits) and “mebibits” (2^20 bits) like the existing byte-based units.

PR #315, opened 2018-09-17
Add support for {Gibi|Mebi|Kilo|etc}bits (#304)

pdobel:
Hi,
I've implemented bit-based units in Information and DataRate classes. Those should help using squants library in telecommunication industry. Issue for that was already open: #304 
Please review and merge if possible.

### DIFF
--- a/shared/src/main/scala/squants/information/DataRate.scala
+++ b/shared/src/main/scala/squants/information/DataRate.scala
@@ -4,7 +4,7 @@ import squants._
 import squants.time.TimeDerivative
 
 /**
-  * Represnets a rate of transfer of information
+  * Represents a rate of transfer of information
   */
 final class DataRate private(val value: Double, val unit: DataRateUnit)
     extends Quantity[DataRate]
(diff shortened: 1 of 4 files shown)


Passarinho4:
@cquiroz @derekmorr  could you review it or merge? It looks good to me.

shared/src/main/scala/squants/information/DataRate.scala
```diff
@@ -32,6 +32,25 @@ final class DataRate private(val value: Double, val unit: DataRateUnit)
...
+  def toGigabitsPerSecond = to(GigabitsPerSecond)
+  def toTerabitsPerSecond = to(TerabitsPerSecond)
+  def toPetabitsPerSecond = to(PetabitsPerSecond)
+  def toExabitsPerSecond = to(ExabitsPerSecond)
+  def toZettabitsPerSecond = to(ZettabitsPerSecond)
+  def toYottabitsPerSecond = to(YottabitsPerSecond)
+
+  def toKibibitsPerSecond = to(KibibitsPerSecond)
+  def toMebibitsPerSecond = to(MebibitsPerSecond)
+  def toGibibitsPerSecond = to(GibibitsPerSecond)
```
cquiroz: These are new to me. Do you have a reference to cite about these units?

shared/src/main/scala/squants/information/Information.scala
```diff
@@ -154,6 +175,91 @@ object Yobibytes extends InformationUnit {
   def symbol = "YiB"
 }
 
+object Bits extends InformationUnit {
+  def conversionFactor = 0.125d
```
cquiroz: Maybe this should be more explicitly set to `1/8`

shared/src/main/scala/squants/information/DataRate.scala
pdobel: Concerning those IEC based units -  as far as i know they are rarely used. Still due to the fact that they were mentioned in https://en.wikipedia.org/wiki/Bit_rate and to be complementary with the added units for Information I decided to include them as well.

shared/src/main/scala/squants/information/Information.scala
pdobel: This can be either `0.125d` or `1d/8` (with double literal or other floating point marking) as both will result in the same value. Originally I've used literal form as it is cleaner in my opinion.

derekmorr:
LGTM

PR #316, opened 2018-09-21
Tiny fixes to improve correctness of predefined currencies

jhnsmth:
I've split the changes into a few commits to ease the review. 

I'm not sure about the last two commits ... 

The third is breaking binary compatibility, maybe we should add some deprecation warning ? 
The fourth one is intended to make the currency in money printed with `Money#toFormattedString` distinguishable.


shared/src/main/scala/squants/market/Money.scala
```diff
@@ -424,30 +424,30 @@ abstract class Currency(val code: String, val name: String, val symbol: String,
 }
 
 object USD extends Currency("USD", "US Dollar", "$", 2)
-object ARS extends Currency("ARS", "Argentinean Peso", "$", 2)
-object AUD extends Currency("AUD", "Australian Dollar", "$", 2)
+object ARS extends Currency("ARS", "Argentinean Peso", "CLP$", 2)
```
cquiroz: This one doesn't seem right. I bet it should be `ARS`

shared/src/main/scala/squants/market/Money.scala
```diff
@@ -424,30 +424,30 @@ abstract class Currency(val code: String, val name: String, val symbol: String,
...
-object RUB extends Currency("RUB", "Russian Ruble", "руб", 2)
+object NZD extends Currency("NZD", "New Zealand Dollar", "NZ$", 2)
+object RUB extends Currency("RUB", "Russian Ruble", "\u20BD", 2)
 object SEK extends Currency("SEK", "Swedish Kroner", "kr", 2)
-object XAG extends Currency("XAG", "Silver", "oz", 4)
-object XAU extends Currency("XAU", "Gold", "oz", 4)
-object BTC extends Currency("BTC", "BitCoin", "B", 15)
+object XAG extends Currency("XAG", "Silver", "Ag", 4)
+object XAU extends Currency("XAU", "Gold", "Au", 4)
+object BTC extends Currency("BTC", "Bitcoin", "B", 15)
```
cquiroz: Is there some unicode for the symbol?

hunterpayne: https://en.bitcoin.it/wiki/Bitcoin_symbol

jhnsmth:
Sorry, for such a long answer... I propose to drop the last commit (just not to mix currency symbol and code, if one wants this its always possible to define, custom Show instance), add BTC symbol and merge. @cquiroz, what do you think ?

garyKeorkunian:
@jhnsmth I agree with your last comment.

jhnsmth:
@garyKeorkunian @cquiroz done! sorry for the delay!

PR #319, opened 2018-11-07
Prepare next version

cquiroz:
Update version number and add some lib updates

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,8 @@ project/plugins/project/
 .history
 .cache
 .lib/
+.bloop/
+.metals/
 
 # Java
 *.class
@@ -52,4 +54,3 @@ tmp/
 # Windows
 Desktop.ini
 Thumbs.db
-
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ branches:
 
 env:
   - TRAVIS_NODE_VERSION="10.9.0" SCALAJS_VERSION="0.6.28"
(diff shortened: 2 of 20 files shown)


garyKeorkunian:
@cquiroz Is this still relevant?  Should we do this in conjunction with the scala 2.13 update when it's ready?

cquiroz:
The Scala 2.13 update is blocked by a compilation issue. Maybe we should do one last release with 2.11/2.12 support and then another including 2.13?

garyKeorkunian:
Sounds good to me.

garyKeorkunian:
Looks like we're having some build issues with the Oracle JDK.  Perhaps something in the travis.yaml needs to be updated.

One thing of interest with our 2.13 issue:  I was at Scala Days in Switzerland in June and learned the Squants was one of the packages the 2.13 team was using to test.  If that's true, it seems unusual that we would be getting this error.  I'll see if I can find out more.

cquiroz:
We need to switch to openjdk 8, the oracle v isn’t supported by Travis anymore

garyKeorkunian:
We were using both.  I found a potential fix, the PR is queued for build now.  If that doesn't work, I'll just yank the OracleJDK.

cquiroz:
I updated the PR to the latest library versions, and did some gymnastics to get scalatest to align across all the build axes. I disabled scala.js `1.0.x` on travis as scalatest is not available for every scala version but we should be able to publish ok

Once approved I can do a release

garyKeorkunian:
LGTM

Issue #320, opened 2018-11-14
Make dimensions available implicitly


vlovgr:
I recently did the work to update to Squants 1.4.0 in Ciris (https://github.com/vlovgr/ciris/pull/190). The changes in https://github.com/typelevel/squants/pull/293 are welcome, but are of limited usefulness because `Dimension`s aren't provided implicitly. In the tests in https://github.com/typelevel/squants/pull/293, you can see the following function:

```scala
def parse[A <: Quantity[A]: Dimension](s: String): Try[A] = {
  implicitly[Dimension[A]].parseString(s)
}
```

which is basically what I want to write for the decoders in Ciris, so that a single definition can support all `Dimension`s. The issue is that `Dimension`s aren't available `implicitly` from Squants, and we can see this in the tests as well, where we have to make them available implicitly.

```scala
implicit val length = Length
implicit val time = Time
implicit val thingee = Thingee
implicit val mass = Mass
```

I'm not sure if there's a reason why `Dimension`s aren't available implicitly, but it would definitely be helpful if they are.

vlovgr:
Fixed by #326 and released in 1.5.0. 🎉 

Here's the update pull request in ciris: https://github.com/vlovgr/ciris/pull/255

cquiroz:
Great to hear

PR #326, opened 2018-11-28
All dimensions were made to have implicit field equal to `this`.

buzden:
This allows us to have nice implicit instance when `Dimension[Q]` is a
companion object to some quantity.

This solves #320 and allows e.g. to write a single cats' `Monoid` (or even `CommutativeGroup`) instance for all quantities (i.e. to have `def groupInstance[Q <: Quantity[Q]: Dimension]: Group[Q]` or so).


cquiroz:
This looks good to me, I'll wait for other committers to chime in before merging

Issue #296, opened 2018-02-05
Money.apply should not use defaultCurrencyMap


veej:
Since now it's possible to define our additional currencies, I expect to be able to create a new `Money` passing a currency code from the additional ones I added, like in the example above:

```
object NMY extends Currency("NMY", "New Money", "$", 2)
implicit val myContext = moneyContext.withAdditionalCurrencies(Set(NMY))

val money = Money(10, "NMY")
```

However, the snippet of code above is throwing an exception, since `Money.apply(amount: BigDecimal, currency: String)` is using a static `defaultCurrencyMap` for the String-to-Currency conversion, instead of the list of currencies from the implicit context.

Issue #323, opened 2018-11-21
NotImplementedError from Money#hashCode


psttf:
with Squants 1.4.0:

```scala
Welcome to Scala 2.12.7 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_181).
Type in expressions for evaluation. Or try :help.

scala> import squants.market._
import squants.market._

scala> Money(123.456, USD)
res0: squants.market.Money = 123.456 USD

scala> res0.hashCode
scala.NotImplementedError: an implementation is missing
  at scala.Predef$.$qmark$qmark$qmark(Predef.scala:284)
  at squants.market.Money$.primaryUnit(Money.scala:404)
  at squants.market.Money$.primaryUnit(Money.scala:385)
  at squants.Quantity.hashCode(Quantity.scala:182)
  ... 36 elided
```

hunterpayne:
This is a real bug.  It is caused by the fact that hashCode comes from the Quantity class.  That implementation uses the hashCode of the Dimension and specifically the hashCode of the primaryUnit.  Because the Money class uses the MoneyContext for the primaryUnit, its lacks an implementation (???) to ensure the MoneyContext implementations are being used.  So the Money class needs its own hashCode implementation which is quite easy.  Here is one possible implementation:

  override def hashCode() = java.util.Objects.hash(currency, amount)

garyKeorkunian:
It looks like this was corrected in #330.  @psttf Can you please verify?

Vrolijkx:
Should be resolved on the latest master. So think we can close this one.

garyKeorkunian:
Resolved by PR #351

PR #330, opened 2018-12-08
Allow currency parsing from implicit context

belerophon:
This PR adds a companion object to `Currency` whose `apply` receives a currency code String, an implicit `MoneyContext` and returns a `Try[Currency]`.

It also changes the methods in `Money` companion that used to parse a currency from a String and relied solely on the `defaultMoneyContext`  to fetch the corresponding `Currency` value:  these methods now receive an implicit `MoneyContext` and use the new `Currency.apply` which, I think, should solve #296.

For my use case, these modifications are useful when working with domain models that specify some `Currency`, that must be parsed from JSON, and when loading `Currency` from typesafe config (using PureConfig).

### DIFF
--- a/shared/src/main/scala/squants/market/Money.scala
+++ b/shared/src/main/scala/squants/market/Money.scala
@@ -387,15 +387,19 @@ object Money extends Dimension[Money] {
   def apply(value: BigDecimal)(implicit fxContext: MoneyContext) = new Money(value)(fxContext.defaultCurrency)
 
   def apply(value: BigDecimal, currency: Currency) = new Money(value)(currency)
-  def apply(value: BigDecimal, currency: String) = new Money(value)(defaultCurrencyMap(currency))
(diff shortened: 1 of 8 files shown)


belerophon:
It looks like I missed this #305 before creating this PR.
Sorry about this.

The comment there:
> I'm not sure about this, passing this as an implicit sounds like an overstretch of implicit. Wouldn't you rather pass it as a plain param?

I agree with @estsauver when he says that the expected usage would be to have the context passed implicitily. Either way, I think the most important point would be to stop using the default context on those operations.

Hope to hear some input on this.
Thanks!

belerophon:
Hi there,
Did someone have the opportunity to take a look into this PR?

psttf:
@garyKeorkunian I've checked the current master (with this PR merged) and it seems like #323 is still a problem.

jchapuis:
Also getting that #323 when running tests

PR #334, opened 2019-01-08
Remove hardcoded dependency on default money context

fredfp:
Squants market package is designed to allow defining a different set of currencies, with optional FX rates.

This flexibility is achieved by using an implicit `MoneyContext` parameter on some methods, with a default value set to `squants.market.defaultMoneyContext`. This way, one can define a new `MoneyContext`, and mark it implicit to have it used instead of the default.

This however isn't consistently implemented throughout the code base. It is especially missing on factory methods of the `Money` object which involve parsing a currency code. As a consequence, one cannot build `Money` instances using custom currencies (i.e., currencies which are not available in the squants market package).

This pull request is meant to fix this issue. It does so by removing any hard-coded references to the default currency set or default money context, and replacing them by an implicitly passed `MoneyContext`.

### DIFF
--- a/shared/src/main/scala/squants/market/Money.scala
+++ b/shared/src/main/scala/squants/market/Money.scala
@@ -387,15 +387,15 @@ object Money extends Dimension[Money] {
   def apply(value: BigDecimal)(implicit fxContext: MoneyContext) = new Money(value)(fxContext.defaultCurrency)
 
   def apply(value: BigDecimal, currency: Currency) = new Money(value)(currency)
-  def apply(value: BigDecimal, currency: String) = new Money(value)(defaultCurrencyMap(currency))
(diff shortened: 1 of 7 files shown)


fredfp:
They do the same except when parsing a `Currency` from a `String`:
- here, https://github.com/typelevel/squants/pull/334, keeps the current squants behaviour i.e., throws when a currency cannot be parsed from a `String`
- https://github.com/typelevel/squants/pull/330 changes the signature of a few `Money.apply` methods to return `Try[Money]` instead of `Money`.

I'd prefer having https://github.com/typelevel/squants/pull/330 merged

garyKeorkunian:
#330 has been approved and will be merged soon.

PR #341, opened 2019-04-11
Closes #324, value field of Information is now Long

### LINKED ISSUE #324: Information addition not always works as expected.
`Bytes(Double.MaxValue / 2.0) == (Bytes(Double.MaxValue / 2.0) + Bytes(1.0))` will evaluate to `true`
This happens because `Information` underlying type is `Double`. Not sure about reasons to choose this, it seems that Long should suit these needs better.

atk91:
Closes https://github.com/typelevel/squants/issues/324


cquiroz:
I understand the motivation of this PR however Double is (for better or worse) in general what we use. A much better solution would be to make squants parametric on the type and make e.g. `Information` use `Long`

garyKeorkunian:
I agree with @cquiroz.  For better or worse that is what it is.  Money is currently the only exception, being a BigDecimal, which was required for that domain.  

I am continuing to work on approaches that will make the type for `value` abstract.

PR #351, opened 2019-09-13
#323 adding hashcode to Money

Vrolijkx:
PR to solve bug #323.
This makes it possible to use Money and Price objects correctly in Sets/Maps. Or when these objects are nested in a case class

### DIFF
--- a/shared/src/main/scala/squants/market/Money.scala
+++ b/shared/src/main/scala/squants/market/Money.scala
@@ -14,6 +14,7 @@ import scala.util.{Failure, Success, Try}
 import scala.language.implicitConversions
 import scala.math.BigDecimal.RoundingMode
 import scala.math.BigDecimal.RoundingMode.RoundingMode
+import java.util.Objects
 
 /**
  * Represents a quantity of Money.
@@ -259,6 +260,12 @@ final class Money private (val amount: BigDecimal)(val currency: Currency)
   }
 
   /**
+   * Override for Quantity.hashCode because Money doesn't contain a primary unit
+   * @return
+   */
+  override def hashCode = java.util.Objects.hash(amount, currency)
+
+  /**
    * Override for Quantity.compare to only work on Moneys of like Currency
    * @param that Money
    * @return Int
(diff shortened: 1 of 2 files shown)


shared/src/main/scala/squants/market/Money.scala
```diff
@@ -14,6 +14,7 @@ import scala.util.{Failure, Success, Try}
 import scala.language.implicitConversions
 import scala.math.BigDecimal.RoundingMode
 import scala.math.BigDecimal.RoundingMode.RoundingMode
+import java.util.Objects
```
cquiroz: This is import is unused and it is breaking the build

shared/src/main/scala/squants/market/Money.scala
```diff
@@ -377,6 +384,9 @@ final class Money private (val amount: BigDecimal)(val currency: Currency)
     * @return Money
     */
   def mapAmount(f: BigDecimal => BigDecimal) = currency(f(amount))
+
+
```
cquiroz: Could you remove the extra lines?

shared/src/main/scala/squants/market/Money.scala
Vrolijkx: I removed the `java.util` package from the hash function.

shared/src/main/scala/squants/market/Money.scala
Vrolijkx: I fixed the formatting back to the original(style).

Vrolijkx:
@cquiroz can somebody merge looks like I don't have the permission to do that.

cquiroz:
@Vrolijkx We've had the policy of having two people approve a PR, however if nobody approves in a few days I'll merge. I was hoping to use your PR to test the automated release but I think I won't be ready by then

shared/src/test/scala/squants/market/MoneySpec.scala
```diff
@@ -151,6 +151,12 @@ class MoneySpec extends FlatSpec with Matchers {
     x != y should be(right = true)
   }
 
+  it should "return consistent hashcode" in {
+    val someMoney = USD(2.1)
+
+    someMoney.hashCode() shouldBe someMoney.hashCode()
```
ybasket: Java requires equal hash codes for equal objects, so I suggest adding a test which also checks against another instance of `USD(2.1)`

Issue #344, opened 2019-06-11
scala 2.13


TimPigden:
Hi, any chance of a 2.13 build please?

cquiroz:
Unfortunately there is an oustanding bug to compile on 2.13 that may take a while to fix

ybasket:
@cquiroz any update on the outstanding bug?

If there's no fix yet, maybe you can give a hint or two about what's the problem or open a WIP PR, so that others can try to help there? I'd gladly invest some time to make squants 2.13 ready if I can.

cquiroz:
We are hitting this bug
https://github.com/scala/bug/issues/11640

ybasket:
The scalac bug was closed today, but it seems as it didn't solve the problem for squants, there's still a cyclic reference error when building against the scalac snapshot `2.13.1-bin-dfde51f` which is supposed to have it fixed (at least if I got the system of snapshot versions right).

Branch to test can be found here: https://github.com/ybasket/squants/tree/scala-2.13

Guess we should try to extract a minimal example of the compilation error and open a new scalac issue…not sure I will soon have the time to do that.

cquiroz:
This is the error in question but I haven't been able to minimize it
```
/Users/cquiroz/Projects/squants/shared/src/main/scala/squants/motion/Acceleration.scala:25:25: illegal cyclic reference involving class Velocity [error] with TimeDerivative[Velocity]
```

PR #363, opened 2019-11-06
2.13: Add Numeric.parseOption implementations

ybasket:
As Scala added `parseOption` to `Numeric` in 2.13 and we define our own `Numeric` instances, we have to add `parseOption` here as well. Implementations are delegating to the existing code to maintain consistency.

Edit: This is part of what's needed for #344.


shared/src/main/scala/squants/AbstractQuantityNumeric.scala
```diff
@@ -37,4 +37,5 @@ abstract class AbstractQuantityNumeric[A <: Quantity[A]](val unit: UnitOfMeasure
   def toFloat(x: A) = x.to(unit).toFloat
   def toDouble(x: A) = x.to(unit)
   def compare(x: A, y: A) = if (x.to(unit) > y.to(unit)) 1 else if (x.to(unit) < y.to(unit)) -1 else 0
+  def parseString(str: String): Option[A] = unit(0).dimension.parseString(str).toOption
```
ybasket: Happy to hear if someone has a more elegant way of getting the dimension from a unit, otherwise I'll add a comment with some explanation of this "hack".

cquiroz: I don't know of another way. A comment would be good

PR #405, opened 2020-07-28
run scalatest rewrites for 3.1

larsrh:
Should unblock #401

This removes uses of deprecated traits and uses the new traits. These are only name changes.


cquiroz:
Thanks a lot

PR #406, opened 2020-07-29
prepare for Dotty

larsrh:
I've verified that `squantsJVM` builds with Dotty 0.26.0-RC1 (tests still pending).

### DIFF
--- a/shared/src/main/scala/squants/Quantity.scala
+++ b/shared/src/main/scala/squants/Quantity.scala
@@ -123,7 +123,7 @@ abstract class Quantity[A <: Quantity[A]] extends Serializable with Ordered[A] {
    * @return Quantity
    */
   def negate: A = unit(-value)
-  def unary_-(): A = negate
+  def unary_- : A = negate
 
   /**
    * Returns the absolute value of this Quantity
@@ -171,7 +171,7 @@ abstract class Quantity[A <: Quantity[A]] extends Serializable with Ordered[A] {
    * @return
    */
   override def equals(that: Any) = that match {
-    case x: Quantity[A] if x.dimension == dimension ⇒ value == x.to(unit)
+    case x: Quantity[_] if x.dimension == dimension ⇒ value == x.asInstanceOf[Quantity[A]].to(unit)
     case _ ⇒ false
   }
 
--- a/shared/src/main/scala/squants/SVector.scala
+++ b/shared/src/main/scala/squants/SVector.scala
@@ -144,7 +144,7 @@ case class DoubleVector(coordinates: Double*) extends SVector[Double] {
 
   type SVectorType = DoubleVector
 
-  def magnitude: Double = math.sqrt(coordinates.toTraversable.map(v ⇒ v * v).sum)
+  def magnitude: Double = math.sqrt(coordinates.map(v ⇒ v * v).sum)
   def angle(coordinateX: Int = 0, coordinateY: Int = 1, unit: AngleUnit = Radians): Angle =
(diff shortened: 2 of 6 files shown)


shared/src/main/scala/squants/Quantity.scala
```diff
@@ -123,7 +123,7 @@ abstract class Quantity[A <: Quantity[A]] extends Serializable with Ordered[A] {
    * @return Quantity
    */
   def negate: A = unit(-value)
-  def unary_-(): A = negate
+  def unary_- : A = negate
```
larsrh, line 126: This looks like it should be a binary-compatible change. If `unary_-` had an empty parameter lists, Dotty would prevent you from calling it as `-x`.

shared/src/main/scala/squants/Quantity.scala
```diff
@@ -171,7 +171,7 @@ abstract class Quantity[A <: Quantity[A]] extends Serializable with Ordered[A] {
    * @return
    */
   override def equals(that: Any) = that match {
-    case x: Quantity[A] if x.dimension == dimension ⇒ value == x.to(unit)
+    case x: Quantity[_] if x.dimension == dimension ⇒ value == x.asInstanceOf[Quantity[A]].to(unit)
```
larsrh, line 174: Dotty prohibits this type assertion (Scala 2 should warn about this). I'm not entirely sure if my "fix" is correct.

shared/src/main/scala/squants/SVector.scala
```diff
@@ -144,7 +144,7 @@ case class DoubleVector(coordinates: Double*) extends SVector[Double] {
 
   type SVectorType = DoubleVector
 
-  def magnitude: Double = math.sqrt(coordinates.toTraversable.map(v ⇒ v * v).sum)
+  def magnitude: Double = math.sqrt(coordinates.map(v ⇒ v * v).sum)
```
larsrh, line 147: Most of the collection conversions here seemed redundant to me, so I deleted them.

shared/src/main/scala/squants/motion/AngularAcceleration.scala
```diff
@@ -3,7 +3,7 @@ package squants.motion
 import squants.mass.MomentOfInertia
 import squants.space._
 import squants.time.{Seconds, Time, TimeDerivative}
-import squants.{AbstractQuantityNumeric, Dimension, Length, PrimaryUnit, Quantity, SiUnit, UnitConverter, UnitOfMeasure}
+import squants.{AbstractQuantityNumeric, Dimension, Length, PrimaryUnit, Quantity, SiUnit, SquantifiedDouble, UnitConverter, UnitOfMeasure}
```
larsrh, line 6: `SquantifiedDouble` is an implicit class that triggers somewhere in this file and I'm surprised that it compiled under Scala 2 (it needs to be in scope).

larsrh:
Tests prove to be difficult. I have another branch where I'm down to 126 errors, but I'm keeping it on hold because it's with an old version of Dotty (0.24). I'd rather wait for ScalaTest to come out for 0.26.

shared/src/main/scala/squants/Quantity.scala
cquiroz: I was planning to put this as a separate PR to be included in the current master. This triggers warnings on 2.13.3

cquiroz:
How did you verify this compiles in dotty?

larsrh:
> How did you verify this compiles in dotty?

I have patched the build definition to include the `sbt-dotty` plugin and fiddled around with some dependencies, then basically:

```
$ sbt
> ++0.26.0-RC1!
> squantsJVM/compile
```

PR #435, opened 2020-12-11
Added Colombian Peso and Polish Zloty in Money supported

captuayo:


cquiroz:
This looks except there is a failing test, Could you check it?

PR #450, opened 2021-03-04
Make `hashCode` of quantities stable across runs

julienrf:
Having a stable `hashCode` can be useful for some things related to serialization.

This is why case objects do have a stable `hashCode`.

Here, I propose to derive the `hashCode` of `Dimension` objects from their fully qualified class names.

### DIFF
--- a/shared/src/main/scala/squants/Dimension.scala
+++ b/shared/src/main/scala/squants/Dimension.scala
@@ -84,6 +84,14 @@ trait Dimension[A <: Quantity[A]] {
   }
 
   implicit val dimensionImplicit: Dimension[A] = this
+
+  override def equals(that: Any): Boolean = that match {
+    case dimension: Dimension[_] => dimension.getClass.getName == this.getClass.getName
+    case _ => false
+  }
+
+  override def hashCode(): Int = getClass.getName.hashCode
+
 }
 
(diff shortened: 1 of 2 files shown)


shared/src/main/scala/squants/Dimension.scala
```diff
@@ -84,6 +84,9 @@ trait Dimension[A <: Quantity[A]] {
   }
 
   implicit val dimensionImplicit: Dimension[A] = this
+
+  override def hashCode(): Int = getClass.getName.hashCode
```
cquiroz, line 88: Isn't there some sort of rule that you should always override `hashCode` and `equals` together?

julienrf: The rule is that two values that are equal must have the same hashCode. This is already the case currently, but maybe I should enforce that?

cquiroz: Right, we could make equals based on FQN or else would be based on object id, the latter should be stable I guess but maybe better to be consistent

cquiroz (approved):
Thanks, looks good to me

julienrf:
Thank you @cquiroz! Do you plan to cut a release with this PR? I see that the other changes since the last release are just library dependency updates.

cquiroz:
Yes I can do a release

PR #454, opened 2021-03-09
Support dotty

cquiroz:
Adds dotty support: Some details:

* scala native support dropped
* scala 2.11 support has been dropped
* documentation in process of being ported to mdoc
* tests can't be run on scala.js due to scalatest lack of support


shared/src/main/scala/squants/SVector.scala
```diff
@@ -80,7 +80,7 @@ trait SVector[A] {
    * @return
    */
   def times(that: Double): SVectorType
-  def * = times _
+  def *(that: Double)  = times(that)
```
cquiroz, line 83: This fails to compile in dotty. I'm not sure if what I'm doing is idiomatic

larsrh: Looks good to me.

shared/src/test/scala/squants/experimental/ExperimentalSpec.scala
```diff
@@ -12,36 +12,38 @@ package squants.experimental
...
 import org.scalatest.flatspec.AnyFlatSpec
 import org.scalatest.matchers.should.Matchers
+// import squants.experimental.DimensionlessConversions._
+// import squants.experimental.SquantsNumeric._
 
 class ExperimentalSpec extends AnyFlatSpec with Matchers {
 
   behavior of "Experimental Code"
 
-  it should "create an Int based Quantity as previously" in {
```
larsrh, line 20: Why are these tests commented out?

shared/src/test/scala/squants/experimental/ExperimentalSpec.scala
```diff
@@ -56,7 +58,7 @@ class ExperimentalSpec extends AnyFlatSpec with Matchers {
   }
 
   it should "add two Quantities that are based on the same numeric type" in {
-    val x = Each(10) + Each(10)
+    val x: Dimensionless[_] = Each(10) + Each(10)
```
larsrh, line 61: This should not be necessary, I think

shared/src/test/scala/squants/experimental/ExperimentalSpec.scala
cquiroz: I had troubles compiling them in dotty, I tried a few variants and gave up, especially considering this is "experimental" code

AtelierFox:
Would love to help if something's still throwing errors, as I want to move a personal project to Scala 3 and this seems like it's almost there. Is there anything on this pull that still isn't working or needs fixing?

cquiroz:
I'll clean up the comments by @larsrh and it just needs to get an approval. We can make a release after merging

cquiroz:
@larsrh Do you think I could get merge this one?

.github/workflows/scala.yml
```diff
@@ -15,6 +15,4 @@ jobs:
           java-version: 1.8
       - name: Run tests
         run: |
-          csbt +squantsJVM/compile +squantsJS/compile
-          csbt squantsNative/compile
```
larsrh, line 19: `squantsNative` should be readded

project/Build.scala
```diff
@@ -7,12 +7,12 @@ import com.typesafe.sbt.osgi.SbtOsgi
 import com.typesafe.sbt.osgi.SbtOsgi.autoImport._
 
 object Versions {
-  val Scala = "2.11.12" // Don't use 2.12 yet to avoid troubles with native
+  val Scala = "3.0.0-RC1"
   val ScalaCross =
-    Seq("2.11.12", "2.12.11", "2.13.3")
+    Seq("2.12.13", "2.13.5", "3.0.0-RC1")
```
larsrh, line 12: Bump to `RC2` to get a suitable ScalaTest

project/Build.scala
```diff
@@ -58,16 +58,11 @@ object Compiler {
     "-encoding", "UTF-8",       // yes, this is 2 args
     "-Xfatal-warnings",
     "-unchecked",
-    "-Xfuture",
-    "-Ywarn-dead-code"
+    // "-Xfuture",
+    // "-Ywarn-dead-code"
```
larsrh, line 62: Remove both

larsrh (changes requested):
The [issue with some tests](https://github.com/typelevel/squants/pull/454#discussion_r591830796) is still open; I recommend either moving them to a Scala 2.x subfolder (I think sbt should do that by default now) or remove the experimental code altogether. Keeping them in but commented is not great.

cquiroz:
I deleted the experimental code, I wanted to keep it but the build started to get very complicated 

let me know if this needs more changes

larsrh (approved):
LGTM!

build.sbt
```diff
@@ -34,27 +33,20 @@ lazy val squants =
   .in(file("."))
   .settings(defaultSettings: _*)
   .jvmConfigure(
-    _.enablePlugins(TutPlugin, SbtOsgi)
+    _.enablePlugins(MdocPlugin, SbtOsgi)
```
julienrf, line 36: By enabling the `MdocPlugin` in the library, this adds the `mdoc` artifacts as dependencies of squants itself. I don’t think this should be the case.

Screenshot from 2021-05-06 16-37-12

Instead, the `MdocPlugin` should be enabled to a separate “documentation” project.

julienrf: I’ve tried to address this problem but I faced other issues. The most important one is that the current README.md does not compile with mdoc.

julienrf: I’ve created #467 to fix this problem.

PR #467, opened 2021-05-07
Enable MdocPlugin on a separate docs module

julienrf:
The goal here is to remove the `mdoc` dependency from the core `squants` artifacts.

I noticed that running mdoc was failing on the current `README.md` file. Unfortunately, this PR does not fix that as well. It seems that the README refers to non-existing APIs.


cquiroz:
Looks good, thanks

PR #471, opened 2021-05-18
Exclude mdoc dependency for downstream apps

martijnblankestijn:
mdocAutoDependency: exclude mdoc as a library dependency as it is not necessary for downstream apps (unless I am mistaken of course ;-) ).

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -56,7 +56,7 @@ lazy val docs =
     .settings(
       scalaVersion := "2.13.6",
       mdocOut := (ThisBuild / baseDirectory).value
-      mdocAutoDependency := false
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -56,6 +56,7 @@ lazy val docs =
     .settings(
       scalaVersion := "2.13.6",
       mdocOut := (ThisBuild / baseDirectory).value
+      mdocAutoDependency := false
```
cquiroz, line 59: It is missing a comma in the previous line

martijnblankestijn: Stupid mistake, thx

PR #509, opened 2022-01-05
feat: Add Density units (k|m|µ|n)g per (k|m|µ|n)l

yangchoo:
Add density units for all combinations of kilo, milli, micro and nano
grams per kilo, milli, micro and nano litres.

Fixed bug with existing multiplication conversion and explicitly typed
all return values for Density class and object.

### DIFF
--- a/shared/src/main/scala/squants/mass/Density.scala
+++ b/shared/src/main/scala/squants/mass/Density.scala
@@ -9,6 +9,9 @@
 package squants.mass
 
 import squants._
+import squants.space.{CubicMeters, Litres, Microlitres, Millilitres, Nanolitres}
+
+import scala.util.Try
 
 /**
  * @author  garyKeorkunian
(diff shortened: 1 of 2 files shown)


shared/src/main/scala/squants/mass/Density.scala
```diff
@@ -19,36 +22,265 @@ import squants._
 final class Density private (val value: Double, val unit: DensityUnit)
     extends Quantity[Density] {
 
-  def dimension = Density
+  def dimension: Dimension[Density] = Density
 
-  def *(that: Volume): Mass = Kilograms(this.value * that.toCubicMeters)
+  def *(that: Volume): Mass = Kilograms(this.toKilogramsPerCubicMeter * that.toCubicMeters)
```
cquiroz, line 28: Is this the fix that you mentioned?

cquiroz (approved):
Thanks a lot for your PR and for including tests

PR #522, opened 2022-03-05
Migrate to sbt-typelevel-ci-release

armanbilge:
So the tricky thing about this is change is it seems that MiMa is enabled for the first time in this project, and there are many many issues.

Apparently master has already broken binary-compatibility compared to the latest release 1.8.3 due to the changes in https://github.com/typelevel/squants/pull/509. This is not considering the 100+ MiMa binary issues if we compare against 1.0.0 :)

One easy strategy may be to declare a clean slate: your next version should be 2.0.0 since it's binary-breaking, and from then on you will preserve binary compatibility (or bump major version again).

WDYT?


cquiroz:
I'm pretty sure binary compatibility has been broken
Bumping to 2.0 and starting with a clean sheet seems the best option

armanbilge:
👍 that's easy enough :)

Btw, if you've been following the discussion, we need to remove the repo secrets from squants and also spire, thanks!
 https://github.com/orgs/typelevel/teams/maintainers/discussions/6?from_comment=2#discussion-6-comment-2

cquiroz:
Yes go ahead, or I can do it if I have the permission

armanbilge:
I don't have permissions, so it will have to be you :)

armanbilge:
Native fails with this:
```
[info] - should return Length when cube rooted *** FAILED ***
[info]   3.0000000000000004 m was not equal to 3.0 m (VolumeSpec.scala:170)
```

armanbilge:
It fails locally for me too ... were you not running this test before?

armanbilge:
Ah, you were only compiling native but not testing.
https://github.com/typelevel/squants/blob/1a8f60808c8d5b9aaa6cef5e6bde26f28edfa19e/.github/workflows/scala.yml#L18

cquiroz:
I think it wasn't possible to compile to native back then, maybe it is now

cquiroz:
Ok, I removed the secrets from both spire an squants

shared/src/test/scala/squants/space/VolumeSpec.scala
```diff
@@ -167,7 +167,7 @@ class VolumeSpec extends AnyFlatSpec with Matchers {
   }
 
   it should "return Length when cube rooted" in {
-    CubicMeters(27).cubeRoot should be(Meters(3))
+    CubicMeters(64).cubeRoot should be(Meters(4))
```
armanbilge, line 170: Easy fix

cquiroz: Ideally you could choose the number type to have exact calculations but squants only supports doubles

armanbilge:
Good to go, but you will need to remove the old status check.

cquiroz (approved):
LGTM, thankks a lot

PR #549, opened 2022-07-27
Cubic feet per X

eugene-sy:
The PR adds support for missing cubic feet per day/minute/second to the existing cubic feet per hour.


cquiroz (approved):
Thanks a lot, sorry for the delay

PR #553, opened 2022-09-15
Improved Money parsing from string literals

lukoyanov:
This is an improvement for the `Money.apply` which tries to parse and create a `Money` from a string literal. 

When a currency being parsed has a high precision (like cryptocurrencies with 18 fractional digits), the current implementation uses `.toDouble` which might not be enough.


lukoyanov:
@cquiroz PTAL.

lukoyanov:
@cquiroz Thank you! Would you please kindly merge this (I don't have write access in this repository).

PR #556, opened 2022-09-19
 Fixed formatDecimals for crypto ETH, BTC, and LTC

lukoyanov:
This is a tiny improvement for those who use `squants` with cryptocurrencies. It fixes format decimal definition for ETH, BTC, and LTC.


cquiroz (approved):
LGTM

PR #559, opened 2022-09-27
Add Ukrainian Hryvnia 🇺🇦 (UAH) currency

VladKopanev:
Add missing UAH currency


cquiroz:
I've always felt we should get the currencies from the JVM db but we can keep them in squants as well

cquiroz (approved):
LGTM

